PR #2, opened 2020-05-27
Handle parameter types and values of any cle function

haesleinhuepf:
Hey @jni ,
CC @pr4deepr 

as just discussed this "morning", I introduced the decorator (I called it "plugin_function"). I'm pretty happy with the result, as it works. there is a little [test script](https://github.com/clEsperanto/pyclesperanto_prototype/commit/d8cbad471f0a0bcf395e91c13e35552296e8e5fa) as well.  

But it feels a bit like I re-invented parts of python. Maybe there is a way to achieve the same thing in a simpler way? Furthermore, I didn't manage to knit in the idea of having npparrays as output as you suggested. Not intenionally, I just think this long block of [if-statements](https://github.com/clEsperanto/pyclesperanto_prototype/commit/6230116906da261677eacf50c4ce989046808ae9#diff-8d78ebb38b47d7962638e2cf9a7269eeR50-R82) should be cleaned up before we introduce even more special cases.... What do you think?

Thanks for your support!

Cheers,
Robert


clesperanto/plugins/add_images_weighted.py
```diff
@@ -1,7 +1,11 @@
 
 from ..core import execute
+from ..core import create_like
+from ..core import plugin_function
+from gputools import OCLArray
 
-def add_images_weighted(input1, input2, output, weight1, weight2):
+@plugin_function(output_creator=create_like)
+def add_images_weighted(input1:OCLArray, input2:OCLArray, output:OCLArray, weight1=1, weight2=1):
```
jni, line 8: Here I would make the types:

```python
Array = Union[np.ndarray, OCLArray]

@plugin_function(output_creator=create_like)
def add_images_weighted(
    image1 : Array,
    image2 : Array,
    *,  # <- this means that later arguments are keyword-only, ie you have to write out=mybuffer
    out : Optional[Array] = None,
    weight1 : float = 1,
    weight2 : float = 1,
):
```

clesperanto/core/plugin_function.py
```diff
@@ -0,0 +1,105 @@
...
+
+            arg_counter += 1
+
+        #print("Got arguments")
+        #print(args)
+        #print("Will pass arguments")
+        #print(target_arguments)
+
+        # execute function with determined arguments
+        return function(*target_arguments)
```
jni, line 92: To match my assumption of some arguments being keyword only, `target_arguments` should be a dictionary and then this would be 
```suggestion
        return function(**target_arguments)
```

clesperanto/core/plugin_function.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+
+
+    @wraps(function)
+    def worker_function(*args, **kwargs):
+        # determine argument spec and default values, values are given as args
+        argument_specification = inspect.getfullargspec(function)
+        defaults_values = argument_specification.defaults
+
+        target_arguments = [None] * len(argument_specification.args)
```
jni, line 34:
```suggestion
        target_arguments = {}  # empty dict
```

clesperanto/core/plugin_function.py
```diff
@@ -0,0 +1,105 @@
...
+                else:
+                    # any other type
+                    target_arguments[arg_counter] = value
+            else:
+                # it wasn't annotated
+                if (value is not None):
+                    # if it's something, hand it over
+                    target_arguments[arg_counter] = value
+                else:
+                    # if it's not set, hand over default values
```
jni, line 77: The whole default counter is unnecessary with the empty dict approach

clesperanto/core/plugin_function.py
```diff
@@ -0,0 +1,105 @@
...
+            if (len(args) > arg_counter):
+                value = args[arg_counter]
+                #print("value")
+                #print(value)
+            else:
+                value = None
+
+            # was the argument annotated?
+            type_annotation = argument_specification.annotations.get(argument);
+            if (type_annotation is not None):
```
jni, line 50: @haesleinhuepf I guess we should skip the type annotation-based dispatch, since the annotation will be *either* NumPy or OCL.

Another point is that we could make `push` itself magical and be a no-op for OCLArrays.

So with all the suggestions I've made so far I think you're getting rid of almost all the if/elses? :joy:

clesperanto/core/plugin_function.py
```diff
@@ -0,0 +1,105 @@
...
+        # determine argument spec and default values, values are given as args
+        argument_specification = inspect.getfullargspec(function)
+        defaults_values = argument_specification.defaults
+
+        target_arguments = [None] * len(argument_specification.args)
+
+        arg_counter = 0
+        default_counter = 0
+        any_input = None
+        for argument in argument_specification.args:
```
jni, line 39: btw, I don't think you need `arg_counter` if you need, but another Python trick:

```python
for arg_counter, argument in enumerate(argument_specification.args):
```

jni:
@haesleinhuepf wanna do another pass with my comments? =)

clesperanto/plugins/add_images_weighted.py
haesleinhuepf: Hey @jni,

"keyword only" means that 
```
add_images_weighted(in1, int2, out, 1, 1)
```
wouldn't work anymore?

clesperanto/core/plugin_function.py
haesleinhuepf: If you mean the arg_counter, I agree :-)

clesperanto/core/plugin_function.py
haesleinhuepf: You might right. [Push(x) a.ka. convert(x, OCLArray)](https://github.com/clij/clij-core/blob/888ab4c5484477d3192bdb37425bdf94a4699ce4/src/main/java/net/haesleinhuepf/clij/CLIJ.java#L449) in clij is also magic. Let's see how many IFs I can spare...

clesperanto/core/plugin_function.py
haesleinhuepf: Great! I should have seen that target_arguments should be a dictionary :-)

clesperanto/core/plugin_function.py
haesleinhuepf: I will never get used to these 🤣

haesleinhuepf:
> @haesleinhuepf wanna do another pass with my comments? =)

Hey @jni 
Yes, please. It's much shorter and concise now. I'm only not sure about the "keyword only" suggestion (see comment above)

Thanks for your support!

clesperanto/plugins/add_images_weighted.py
lagru: Yes.

haesleinhuepf: Hey @lagru 👋 

pitty. That would mean that [this](https://clesperanto.github.io/images/clijmapy.mp4) wouldn't work - the major reason of the whole exercise. ;-)

So I guess we have to live without that feature. At least on this abstraction layer....

jni: Which feature are you talking about exactly? Are you saying you are lazy and don't want to tell me your parameter names to make sure that they are clear to read? ;)

clesperanto/core/plugin_function.py
```diff
@@ -0,0 +1,79 @@
...
+                    # if not set and should be an image, create an image
+                    # create a new output image with specified/default creator
+                    target_arguments.update({argument: output_creator(any_ocl_input)})
+                else:
+                    # if it's not set and should be something else than an image, hand over default values
+                    if (len(defaults_values) > default_counter):
+                        target_arguments.update({argument: defaults_values[default_counter]})
+                        default_counter += 1
+                    else:
+                        target_arguments.update({argument: None})
```
jni, line 68: Why are you filling in defaults at all? When you are using a dict, calling `function(**target_arguments)` will fill in the arguments in `target_arguments` and leave the rest as defaults.

clesperanto/core/push.py
```diff
@@ -25,5 +28,8 @@ def push(nparray):
     return temp2
 
 def push_zyx(nparray):
```
jni, line 30: I'd rename this `nparray` argument to `anyarray` or such.

clesperanto/core/plugin_function.py
```diff
@@ -0,0 +1,79 @@
...
+        any_ocl_input = None
+
+        for arg_counter, argument in enumerate(argument_specification.args):
+            #print("---\nparsing argument " + argument)
+            if (len(args) > arg_counter):
+                value = args[arg_counter]
+            else:
+                value = None
+
+            if (isImage(value)):
```
jni, line 47: `is_image` ;)

clesperanto/plugins/add_images_weighted.py
haesleinhuepf: With "feature" I mean specifying parameter names is optional. 

This needs to work (Reason: Make the code run in other languages. clEsperanto)
```
add_images_weighted(in1, int2, out, 1, 1)
```

and this is optional:
```
add_images_weighted(in1, int2, out, weight1=1, weight2=1)
```

Is there any secret pythonic rule I violate with that?

clesperanto/core/plugin_function.py
haesleinhuepf: I was thinking of that while sleeping in yesterday 🤣

clesperanto/core/plugin_function.py
```diff
@@ -0,0 +1,79 @@
...
+
+        default_counter = 0
+        any_ocl_input = None
+
+        for arg_counter, argument in enumerate(argument_specification.args):
+            #print("---\nparsing argument " + argument)
+            if (len(args) > arg_counter):
+                value = args[arg_counter]
+            else:
+                value = None
```
jni, line 45: Are we missing kwargs here?

clesperanto/core/plugin_function.py
haesleinhuepf: Right! I didn't see that. But it makes sense 👍

clesperanto/core/plugin_function.py
```diff
@@ -0,0 +1,79 @@
...
+            # default: keep value
+            target_arguments.update({argument: value})
+
+            # was the argument annotated?
+            type_annotation = argument_specification.annotations.get(argument);
+            if (value is None):
+                if (type_annotation is Image):
+                    # if not set and should be an image, create an image
+                    # create a new output image with specified/default creator
+                    target_arguments.update({argument: output_creator(any_ocl_input)})
```
jni, line 61: When adding just one key, it is more common to write:
```suggestion
                    target_arguments[argument] = output_creator(any_ocl_input)
```

clesperanto/core/plugin_function.py
```diff
@@ -0,0 +1,79 @@
...
+            else:
+                value = None
+
+            if (isImage(value)):
+                value = push(value)
+                # value is for sure OpenCL, we keep it in case we have to create another one of the same size
+                any_ocl_input = value
+
+            # default: keep value
+            target_arguments.update({argument: value})
```
jni, line 53: Like below:
```suggestion
            target_arguments[argument] = value
```

clesperanto/plugins/add_images_weighted.py
```diff
@@ -1,7 +1,11 @@
 
 from ..core import execute
+from ..core import create_like
+from ..core import plugin_function
+from ..core import Image
 
-def add_images_weighted(input1, input2, output, weight1, weight2):
+@plugin_function(output_creator=create_like)
+def add_images_weighted(input1:Image, input2:Image, output:Image, weight1:float=1, weight2:float=1):
```
jni, line 8:
```suggestion
def add_images_weighted(input1:Image, input2:Image, output :Image = None, weight1:float=1, weight2:float=1):
```

clesperanto/plugins/add_images_weighted.py
jni: It's just good practice in "modern" Python to say that the names should always be there. It's easy enough to strip the names out in other languages? Other languages will probably use the name `addImagesWeighted` anyway and so on, no?

jni:
@haesleinhuepf I've done one more round. We don't have to decide on keyword-only arguments in this PR. =) But there's still some simplification that can be done here. It's exciting! Does the test pass currently btw???

clesperanto/plugins/add_images_weighted.py
haesleinhuepf: Java-esque languages (Java, Matlab, JavaScript) don't like the `method(parameter=value)` pattern afaik. To circumvent that, developers prefer the builder pattern `method().setParameter(value).run()`. Which is nice but also complicated, especially for coding newbies. 
 
I'd like to keep things simple - and copy-pastable between languages. Thus, all  users from all languages have to step a little step back and use very simple method calls.

Btw. the JavaScript recorder records something like
```
image1 = ...;
image2 = ....;
output = ....;
weight1 = 1;
weight2 = 1;
add_images_weighted(image1, image2, output, weight1, weight2);
```
And this should run ins all target languages. 

I suggest to leave it as it is for now. But any idea/suggestion is welcome!

clesperanto/core/plugin_function.py
haesleinhuepf: ood catch!
 I'm combining all parameters in kwargs now. This obviously includes the given ones

haesleinhuepf:
Hey @jni,

ok, I think we're through. I would merge this and [another branch](https://github.com/clEsperanto/pyclesperanto_prototype/tree/more_ops) where I just translated many operations. I can then build workflows and see if everything works together and is stable. That's the interesting part. Looking forward to see how the "garbage collector" of pyopencl performs ;-)

> Does the test pass currently btw???

Tests passed even if they shouldn't. I just fixed that and added more tests.

If you see any other issue, let me know. Otherwise I would merge next week.

Thanks again, this was/is very helpful!

Cheers,
Robert

clesperanto/plugins/add_images_weighted.py
jni: I honestly think that copy-pasting code between languages should not be a goal, only making the code recognisable between languages. When in Rome, do as the Romans do, and idiomatic Python is different from idiomatic Java. Completely unifying their syntax is, in my opinion, a recipe for a compromise where everyone is unhappy. Not only that, but people seeing code like this when they are learning a language will then learn antipatterns for that language.

As mentioned elsewhere, we can postpone this decision, that's fine.

jni:
@haesleinhuepf please see #4 before merging this. =)

haesleinhuepf:
Thanks again everyone! I'll add some more operations, benchmarks and then we'll see how this develops.

Cheers,
Robert

PR #4, opened 2020-05-31
Some final touches to handle parameter types PR

jni:
@haesleinhuepf I've made a few extra changes that I think were too nitpicky to just comment on.

There's two more changes that need to be handled, I'll try to get to them but commenting here in case I don't and you get itchy merge fingers. =)

- Currently, if the user passes in arguments as keywords, *they won't get validated at all*. That's a problem. Example: `add_images_weighted_2d(image1=my_numpy_array, image2=my_numpy_array2)` will fail because `*args` is empty and `**kwargs` is never iterated on to ensure correct types.
- Once you do that, you still have the issue that we want to pull if `out=[some numpy array]`.
- Finally, you should use pytest for testing, rather than a standalone file to run. To do this, you just put your tests in files starting with `test_`, and each test is a function starting with `test_`. I wrote a guide about this a long time ago [here](https://ilovesymposia.com/2014/10/01/continuous-integration-0-automated-tests-with-pytest/).

For this PR, I tried to keep each commit self-contained, I suggest looking at each individually.

I'll try to address those issues mentioned above but feel free to take over if I haven't when you wake up. =)

### DIFF
--- a/clesperanto/core/plugin_function.py
+++ b/clesperanto/core/plugin_function.py
@@ -47,32 +47,32 @@ def plugin_function(
 
         for arg_counter, argument in enumerate(argument_specification.args):
             #print("---\nparsing argument " + argument)
-            if (len(args) > arg_counter):
+            if arg_counter < len(args):
                 value = args[arg_counter]
             else:
                 value = None
 
-            if (is_image(value)):
+            if is_image(value):
                 value = push(value)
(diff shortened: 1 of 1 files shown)


haesleinhuepf:
Awesome @jni ! Thanks - especially for the links. Will produce more pythonesque docstrings in the future 💪

PR #8, opened 2020-06-03
apply_threshold operation

pr4deepr:
Added the apply_threshold operation


pr4deepr:
@haesleinhuepf , could you have a look at this?

haesleinhuepf:
Hey @pr4deepr ,

awesome, this looks almost perfect! Just wondering: applyThreshold is no method in CLIJ2 - where did you get the OpenCL kernel from? Furthermore, how is it different from [greater or equal](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/clesperanto/plugins/greater_or_equal.py)? I'm just wondering; if you may have written an operation that exists already :-(

But the benchmarking on the 2080 looks very interesting! Is it a 2080 or a 2080 TI?

Cheers,
Robert

pr4deepr:
Oooh, I realised I made a few errors.. :)
Soo, I got the OpenCL kernel from the master branch, not the development branch.
Just to confirm, all the kernels you are importing here are from the development branch, right?
I can see the documentation as well for the placeholders.. Thanks..
The greater_constant one works exactly like the apply_threshold, but is the documentation right? It says:
"""Determines if two images A and B greater pixel wise. 
      f(a, b) = 1 if a > b; 0 otherwise. 
Isn't that the same documentation as greater or equal?
Should it be:
"""Determines if images A greater than scalar s pixel wise (similar to threshold with value s) 
Returns a binary image
or something similar?
Well, duplicated effort, but I've learned a fair bit..

***
The graphics card is a 2080 OC actually. 
I've also done a benchmarking on an NVIDIA MX150 (2GB, 384 cores) on my laptop. [LINK](https://github.com/pr4deepr/pyclesperanto_prototype/blob/master/benchmarking_NVIDIA_MX150.pdf)

pr4deepr:
Question/suggestions
- Do you think its a good idea to provide environment.yml in the github webpage to enable easy configuration of a conda environment for pyclesperanto? 
- How about migrating all the benchmarking pdfs into a folder?

haesleinhuepf:
> The greater_constant one works exactly like the apply_threshold, but is the documentation right?

Ah you're right. I need to fix the documentation. Btw. the documentation in clEsperanto is automatically generated. There is a [Java "script"](https://github.com/clij/clij2-code-generators/blob/master/src/test/java/net/haesleinhuepf/clijx/codegenerator/PyCleDocumentationInjector.java) that takes the documentation from CLIJ2 and inputs it here in this repository. Without code-generators like this, it would be just impossible to maintain code and documentation for 300 operations ;-)
Just saying, if you write a plugin here, it's documentation may be automatically overwritten. So better don't write documentation yet.

> Question/suggestions

These are absolutely valid points. Not high priority as this is a protoype repository, but yes, we should do this. 

Btw. I mentioned in our last zoom meeting that this prototype may be discontinued at some point. This gives us the opportunity to work a bit messy while learning something and heading towards a proper solution. The development approach is called [rapid prototyping](https://en.wikipedia.org/wiki/Rapid_application_development). So don't worry, if we code something here that's awesome, we will take it with us to the next prototype. But by calling it a prototype we communicate that we have the option of dropping major parts and trying something different in a future prototype ;-)

Btw. if you want to translate a CLIJ2 operation to clEsperanto, for example [Crop](https://github.com/clij/clij2/blob/master/src/main/java/net/haesleinhuepf/clij2/plugins/Crop2D.java#L37) is still on the todo list and actually quite similar to [paste](https://github.com/clij/clij2/blob/master/src/main/java/net/haesleinhuepf/clij2/plugins/Paste2D.java#L36) aka [paste](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/clesperanto/plugins/paste.py).

Cheers,
Robert

pr4deepr:
Good to know about the documentation generator. Looks like there are a few options in Python. 

BTW, I am having a bit of trouble with the Crop CLIJ2 operation even though you gave me the examples. My understanding is with Crop operation you pass the desired x,y coordinates of the image with the width and the height required.
How do you go about incorporating the width and the height into the opencl kernel crop_2d_x.cl?  Is it similar to the [draw_box_2d](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/clesperanto/plugins/draw_box_2d_x.cl)?
OR
Do you create an output image of desired height and width, and then draw the src image onto this from user-specified x,y coordinates? if this makes any sense?

haesleinhuepf:
> Do you create an output image of desired height and width, and then draw the src image onto this from user-specified x,y coordinates? if this makes any sense?

Yes! [That's](https://github.com/clij/clij2/blob/master/src/main/java/net/haesleinhuepf/clij2/plugins/Crop2D.java#L62) the way to go. In python, we need to reinvent this procedure a bit. Crop is a special case, the output image is not of the same size as the input image. That's why create_like is not the way to go. But for the moment, you could maybe assume that the output image exists in advance and has the right size...

haesleinhuepf:
Hey @pr4deepr,

I will close this PR for now. If you want to merge your magic gui at some point (we might want to have it in a separate repository though), feel free to file a new PR.

Thanks again. The GUI is greate work!

Cheers,
Robert

PR #13, opened 2020-07-15
some install ideas

tlambert03:
I see we were doing the same thing at the same time... looks like you've already got a reorg branch going on.  So you probably won't want to merge this, but I'll leave it for you to look through if you want.  Couple things to point out:

- you'll want to add a `install_requires` section to your setup.py ... otherwise the user won't get your dependencies.
- this PR vendors the couple parts of GPUtools that you use, and honestly, I think you'll probably want to just own those bits anyway, so this just depends on numpy and pyopencl (and toolz cause of your one curry decorator).
- that `package_data` section in setup.py is confusing, and doesn't really do what it seems like it should.  so your kernels aren't actually getting packaged in the current PyPI release.  See the `MANIFEST.in` file here.
- for iterating on real PyPI without having to keep using up version strings, you can add `rc` to your string (e.g. `0.0.4rc1`) and then install with `pip install --pre` to test it out.  Or you can use [TestPyPI](https://packaging.python.org/guides/using-testpypi/)
- a good tip is to build locally with `python setup.py sdist` and then examine the files in your  `/dist` folder to make sure everything made it in ([`check-manifest`](https://pypi.org/project/check-manifest/) is also very handy here)


haesleinhuepf:
Hey @tlambert03 ,

> * you'll want to add a `install_requires` section to your setup.py ... otherwise the user won't get your dependencies.

Great, I couldn't figure this one out!

> * this PR vendors the couple parts of GPUtools that you use, and honestly, I think you'll probably want to just own those bits anyway, so this just depends on numpy and pyopencl (and toolz cause of your one curry decorator).

awesome, you're even reading my thoughts! Yes! Cool. Pyopencl-installation remains a hurdle on windows, but this makes it simpler! And it answers some questions about pyopencl I had in mind.

> * that `package_data` section in setup.py is confusing

I copied it from gputools. But your solution looks better!

Ok, cool! Thanks for all your hints!

Shall I merge that right away or do you still want to add stuff?

Cheers,
Robert

tlambert03:
> > that package_data section in setup.py is confusing
> I copied it from gputools. But your solution looks better!

to be clear, I didn't mean that _yours_ was confusing ... I meant that the whole setting is confusing.  https://stackoverflow.com/a/14159430/1631624

> Shall I merge that right away or do you still want to add stuff?

I guess this depends on how much of this you want vs your other reorg branch.  Do you like the tier 0/1 reorg for other reasons?  or was that an attempt to get the files to PyPI?  We should also merge some of the other niceties of your setup.py file over there (like readme) ... unless you want to just start with this and take it from there?

haesleinhuepf:
> to be clear, I didn't mean that yours was confusing ... I meant that the whole setting is confusing.

Yes. Sure. All good :-)

I think, I would merge your branch into mine; check some things out with the tiers and tests folders and then merge into master 👍

Thanks again!

tlambert03:
ok, go for it!  Once it's in master I'll take another look and maybe setup github deployments (if you want) and/or pytest.

haesleinhuepf:
closing this because changes are also in #19

PR #14, opened 2020-07-15
cleanup execute function

tlambert03:
This PR uses some syntactic sugar to cleanup the main execute helper a bit.
Note however, that it also changes the function signature, and therefore would require a change to all of the `plugins/*.py` files (Which I'm also happy to do if you like this pattern).  The new call looks like:
```
execute("add_image_and_scalar", output.shape, parameters)
```
obviously, this leans heavily on your naming conventions (specifically: that kernels and filenames are the same, and that the nD is in the kernel name). If you know of exceptions to that convention let me know, and this PR could also include a test to assert that convention using static analysis.

This is partially a step towards auto-generated python files, which I'm still relatively confident should be possible here.  I started playing a bit with [`pycparser`](https://github.com/eliben/pycparser) which works to pull the kernel name and signature params from your kernel files.  So, also curious to hear if you have particular doubts/concerns about that possibility


haesleinhuepf:
> obviously, this leans heavily on your naming conventions

Yes, the intention of the yet-not-documented naming convention was to simplify exactly that! So I'll keep this PR, but it might take some time until we actually can apply it.

> So, also curious to hear if you have particular doubts/concerns about that possibility

Regarding the auto-generation of the .py-files. I started discussing this with @pr4deepr and @jni earlier. To make that clear: I'm not against it, but I shall have warned: I assume the effort will explode while doing it, because almost every single .py file is a special case IMHO. Let me explain:
* There are about 350 operations in CLIJ, but less than 150 actually call OpenCL (I just created the terminology, they are "tier1"). Thus, we can at max generate half of the .py files.
* The differences between the individual tier1.py files are tiny. Examples: 
  * If you want to [crop](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/clesperanto/plugins/crop.py#L32) an image, the global size of the OpenCL-kernel corresponds to the target image (`output.shape`) as it describes the size of the cropped region. In case of [paste](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/clesperanto/plugins/paste.py#L21), the global size of the kernel comes from the input image size as it describes the size of the part to be inserted into the target image (`src.shape`).
  * Some .py files [don't take care](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/clesperanto/plugins/absolute.py#L23-L38) of image dimensionality. Others [have to](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/clesperanto/plugins/maximum_sphere.py#L21). 
  * There are [separable](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/clesperanto/plugins/mean_box.py#L14) filters and [not-separable](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/clesperanto/plugins/mean_sphere.py) which are invoked quite differently. 
  * Most kernels have input image size and output image size identical. In other cases they don't (trivial example: maximum-projection, complex example: histogram) Generation of output-image size belong in the python files, as its operation specific, I just haven't managed yet to find a good formalism.
* I think it would cost several weeks for find a way of formal specification (in a new domain specific language) to specify how these .py-files could be generated. And it would 95% treat special cases. At the end, one would need to write the specifications for the auto-generator which takes as long as writing the .py-files. 

Thus, I went down an intermediate route: Parts of the .py files are [auto-generated, namely the doc-strings](https://github.com/clij/clij2-code-generators/blob/master/src/test/java/net/haesleinhuepf/clijx/codegenerator/PyCleDocumentationInjector.java). That's quite important, because I would like to keep things synced with the Java-counter part and not write documentation twice. At the moment, the generator reads from java and writes to python, I would love to read from a third source (test based, human readable) and write to java, python, macro, matlab, javascript, ...

There are some related sub-projects which are more worthwhile to dig into IMHO:
* Making the kernels process nD images is tricky. Distribution of task a.k.a. allocating compute cores to parts of the memory to minimize clashes is hardware-optimized for 3D. OpenCL doesn't even have methods for reading pixels in 4D-images.
* Translate [tier0](https://github.com/clEsperanto/pyclesperanto_prototype/tree/tlambert-forks-setup.py/pyclesperanto_prototype/tier0) to CUDA without harming the API used in [tier1](https://github.com/clEsperanto/pyclesperanto_prototype/tree/tlambert-forks-setup.py/pyclesperanto_prototype/tier1). For that, every cl-file needs to be translated to a cu-file - THIS should be doable automatically ;-) 🤯

tlambert03:
> Yes, the intention of the yet-not-documented naming convention was to simplify exactly that! So I'll keep this PR, but it might take some time until we actually can apply it.

the core of this PR is really more about improving the body of that execute function,  things like caching previously read kernels, and generally simplifying the defines.  I can revert the part of this PR that changes the signature of `execute` if you want, and then when a test is written to assert the naming convention, simplify that signature and change all the `plugin/*.py` files

> There are about 350 operations in CLIJ, but less than 150 actually call OpenCL (I just created the terminology, they are "tier1"). Thus, we can at max generate half of the .py files.

I'm not proposing that you can autogenerate the entire CLIJ API... just that the python API into the kernels themselves can (and should) be autogenerated.  And of course if you need to compose functions and/or add specific logic, then of course, that can't be entirely autogenerated.  but even there, we could create some decorators that could handle certain stereotyped dimension-handling issues (like in `maximum_sphere`).

> I think it would cost several weeks for find a way of formal specification (in a new domain specific language) to specify how these .py-files could be generated. And it would 95% treat special cases. At the end, one would need to write the specifications for the auto-generator which takes as long as writing the .py-files.

I agree that there will always be exceptions, such as separable kernels, but I guess I'm not quite as pessimistic about the amount of time required vs the amount of time (and potential bugs) saved.

> That's quite important, because I would like to keep things synced with the Java-counter part and not write documentation twice. At the moment, the generator reads from java and writes to python, I would love to read from a third source (text based, human readable) and write to java, python, macro, matlab, javascript, ...

I get it, definitely shouldn't be duplicating anything.  I guess I was looking at this as a big collection of kernels, and a little less as a porting of a Java project to python.  But you should certainly stick with whatever pipelines feel right for you, and we'll all be very appreciative regardless 😄 
I was imagining that the documentation would live in the CL files themselves, since that seems to be the one constant across all the target languages, right?

clesperanto/core/execute.py
```diff
@@ -1,125 +1,108 @@
...
+                    f"#define IMAGE_{key}_PIXEL_TYPE float",
+                    f"#define IMAGE_SIZE_{key}_WIDTH {width}",
+                    f"#define IMAGE_SIZE_{key}_HEIGHT {height}",
+                    f"#define IMAGE_SIZE_{key}_DEPTH {depth}",
+                    f"#define POS_{key}_TYPE {pos_type}",
+                    f"#define POS_{key}_INSTANCE(pos0,pos1,pos2,pos3) ({pos_type}){pos}",
+                    f"#define READ_{key}_IMAGE(a,b,c) read_buffer{img_dims}d{typeId}"  # !
+                    "(GET_IMAGE_WIDTH(a),GET_IMAGE_HEIGHT(a),GET_IMAGE_DEPTH(a),a,b,c)",
+                    f"#define WRITE_{key}_IMAGE(a,b,c) write_buffer{img_dims}d{typeId}"  # !
+                    "(GET_IMAGE_WIDTH(a),GET_IMAGE_HEIGHT(a),GET_IMAGE_DEPTH(a),a,b,c)",
```
jni, line 86: I feel like this should be a string, e.g.

```python
IMAGE_HEADER = """
#define CONVERT_{key}_PIXEL_TYPE clij_convert_float_sat
#define IMAGE_{key}_TYPE __global float*
...
"""
```

Then this would be 

```python
defines.extend(
    IMAGE_HEADER.format(**locals()).split('\n')
)

tlambert03:
ok, @haesleinhuepf, I reverted the change to the `execute()` signature, so this can be merged without any assertions about your naming conventions, should you be so inclined...

clesperanto/core/execute.py
tlambert03: done

haesleinhuepf:
> I'm not proposing that you can autogenerate the entire CLIJ API... just that the python API into the kernels themselves can (and should) be autogenerated.

Yes. That's tier0. We didn't have these categories in CLIJ. I must also admit I got a bit lost in all these kernels and operations. Did you know that I basically [rewrote almost all of these OpenCL-files when going from clij to clij2](https://github.com/clEsperanto/clij-opencl-kernels/commit/9546d0f187f1111e2e8a5215156e2d5f6930cfaa)? 😣 😅 

> I agree that there will always be exceptions, such as separable kernels, but I guess I'm not quite as pessimistic about the amount of time required vs the amount of time (and potential bugs) saved.

I will actively support you if you want to give it a go. It would be a very useful feature on a broader picture. CLIJ and clEsperanto were always thought as collections of opencl-kernels and related operations. If there also some tools for dealing with many opencl filters, even better. Maybe the code generator can also generate Java, C++ and Julia? @VolkerH might also be interested in this topic.

> I was imagining that the documentation would live in the CL files themselves, since that seems to be the one constant across all the target languages, right?

Yes. Great idea! Sounds to me like [doxygen](https://www.doxygen.nl/index.html) - any other opinions?

tlambert03:
> Yes. Great idea! Sounds to me like doxygen - any other opinions?

documenting C code is new territory for me!  doxygen seems like as good as any

Issue #15, opened 2020-07-15
todo list: operation translation


haesleinhuepf:
I'm just trying to get organized and split all CLIJ2 operations (which might be translated to clesperanto at some point) into tiers: 
* tier0 is the core level (push, pull, execute,...), 
* tier1 are the actual OpenCL wrappers, 
* tier2 doesn't call OpenCL directly but combines operations in tier1 to do useful stuff. 
The higher the tier, the more abstract from the GPU it is. Methods in tier 4 cannot be implemented and/or tested without certain methods in lower tiers. Thus, tiers define the order of when to translate them.
Very high tier operations may even have additional dependencies such as scikit-image, or may need to be implemented on the CPU, e.g. with numpy or scipy and may thus live in a different repository at some point...

## Tier 0: core functions
- [ ] Clear
- [ ] ClInfo
- [x] Create2D
- [x] Create3D
- [x] Execute
- [x] ExecuteSeparableKernel
- [ ] GetGPUProperties
- [ ] GetDimensions
- [ ] InvalidateKernelCache
- [ ] ListAvailableGPUs
- [ ] Print
- [x] Pull
- [ ] PullString
- [x] Push
- [ ] PushString
- [ ] Release
- [ ] ReportMemory

## Tier 1: OpenCL Wrappers
- [x] Absolute
- [x] AddImageAndScalar
- [x] AddImages
- [x] AddImagesWeighted
- [ ] AffineTransform
- [ ] AffineTransform2D
- [ ] AffineTransform3D
- [ ] ApplyVectorField2D
- [ ] ApplyVectorField3D
- [ ] ArgMaximumZProjection
- [x] BinaryAnd
- [x] BinaryEdgeDetection
- [x] BinaryFillHoles
- [x] BinaryNot
- [x] BinaryOr
- [x] BinarySubtract
- [x] BinaryXOr
- [x] Convolve
- [x] Copy
- [x] CopySlice
- [ ] CountNonZeroPixels
- [ ] CountNonZeroPixels2DSphere
- [ ] CountNonZeroPixelsSliceBySliceSphere
- [ ] CountNonZeroVoxels3DSphere
- [ ] CountTouchingNeighbors
- [x] Crop2D
- [x] Crop3D
- [ ] DetectLabelEdges
- [ ] DetectMaxima2DBox
- [ ] DetectMaxima3DBox
- [ ] DetectMaximaBox
- [ ] DetectMaximaSliceBySliceBox
[...]

Issue #6, opened 2020-06-02
Use pytest as testing framework


jni:
In #2, the "tests" are a file to be run with Python. [pytest](https://docs.pytest.org/en/latest/) provides a lot of nice features like running *all* test files, specifying which tests exactly are failing, and much more. To do so, tests should be placed in files called `test_something.py`, and within each file, each test should be its own function, `def test_something_specific():`, which can have assertions etc within it.

haesleinhuepf:
Just adding this link from our earlier discussion:
https://ilovesymposia.com/2014/10/01/continuous-integration-0-automated-tests-with-pytest/
:-)

PR #19, opened 2020-07-16
Deployment, tests and tiers

haesleinhuepf:
Hey @jni, @lagru, @pr4deepr and @tlambert03,

with the generous support by @tlambert03 I managed to do some major breaking changes:
* We got rid of the gputools dependency by copying over the important core stuff to [_pycl.py](https://github.com/clEsperanto/pyclesperanto_prototype/blob/tlambert-forks-setup.py/pyclesperanto_prototype/tier0/_pycl.py). That gives us more freedom and independence. Side note: As mentioned earlier, there is a C++ backend in the making in Paris. I'm expecting it to appear here in summer. It would allow us to also get rid of the pyopencl dependency. I have really no opinion if this is a good or bad idea.
* We reorganized [tests](https://github.com/clEsperanto/pyclesperanto_prototype/tree/tlambert-forks-setup.py/tests). I'd say there are still hundreds of tests missing and heavy work necessary, but at least we are #6 using pytest now.
* I renamed the whole thing from "clesperanto" to "pyclesperanto_prototype" to clearly communicate that this is experimental stuff. I would like to prevent that workflow builders now start adopting it and later complain because we massively break the API and backwards compatibility.
* We managed to setup deployment to pypi (was my first deploy to this repository 🥳 ). It was suprisingly easy.
* I split core and plugins into "tiers". Background is: Within clij there are a lot of dependencies which are not visible on the first view - all 350 operations lie in one folder. I wanted to improve on this to get a better overview. For example: the `top-hat` filter (tier2) needs the `minimum` and the `maximum` filters (tier1). These two need `execute` (tier0). From that a hierarchy builds up autonomously. I listed all operations and their corresponding tiers [here](https://github.com/clEsperanto/pyclesperanto_prototype/issues/15). I will slowly and steadly continue translating filters, from tier0 to tier8/9. Especially the last ones become interesting. That might be wrappers towards numpy, scipy and scikit-image. Not sure yet how to do all this. But there's plenty of time.
* If I want to prevent that users access tiers, it is a good practice to call these packages "_tier" with a leading underscore, or?
* I also added a LICENSE file saying "BSD" and contains also your names. Let me know if this is not ok.

Btw.: @tlambert03 : I invited you to the organisation. Join if you like! :-)

I know there was tons of stuff changed and there's still tons of todos. Anyway: any feedback is welcome!

Cheers,
Robert


tlambert03:
> with the generous support by @tlambert03 I managed to do some major breaking changes:

lol.  always happy to help people break stuff. 😂 

> We reorganized tests. I'd say there are still hundreds of tests missing and heavy work necessary, but at least we are #6 using pytest now.

Once we have the basic pytest setup in there, I think it might be nice to setup a list of `[kernel_name, input, expected output]` and then run them _all_ with a single parameterized pytest test.  sortof a smoke test that becomes really easy to add to: https://docs.pytest.org/en/stable/parametrize.html

> I listed all operations and their corresponding tiers [here](https://github.com/clEsperanto/pyclesperanto_prototype/issues/15).

super interesting and helpful to see!

> If I want to prevent that users access tiers, it is a good practice to call these packages "_tier" with a leading underscore, or?

yep, though there's no strict "private" things in python, that's the convention, and the implication is that you can change that stuff without requiring a minor version bump.  but adherence to that varies I think.

PR #20, opened 2020-07-16
cleanup transpose

tlambert03:
for 2D and 3D arrays, the `swapaxes` calls here are equivalent to [`numpy.ndarray.T`](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.T.html)

I did want to ask though... why is it that everything is being transposed before and after transfer to/from the GPU?  Is this an assumption made by the kernels?  I guess due to ZYX convention in numpy and (presumably) XYZ in Java?

### DIFF
--- a/pyclesperanto_prototype/_tier0/create.py
+++ b/pyclesperanto_prototype/_tier0/create.py
@@ -1,26 +1,25 @@
 from ._pycl import OCLArray
 import numpy as np
 
+
 def create(dimensions):
 
-    '''
+    """
     Convenience method for creating images on the GPU. This method basically does the same as in CLIJ:
 
     https://github.com/clij/clij2/blob/master/src/main/java/net/haesleinhuepf/clij2/CLIJ2.java#L156
 
     :param dimensions: size of the image
     :return: OCLArray, potentially with random values
-    '''
-    if isinstance(dimensions, OCLArray):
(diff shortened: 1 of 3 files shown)


haesleinhuepf:
> I guess due to ZYX convention in numpy and (presumably) XYZ in Java?

Yes exactly. I'm considering having `push_xyz` and `push_zyx` methods on all platforms. Even though it appears weird. And a good self-explanatory naming scheme would be nice. I had similar issues with matlab. 

Fun fact: It should actually matter performance wise if you process an image stack of size 1024x1024x64 or 64x1024x1024... But I haven't digged into this yet...

tlambert03:
note also: 
```python
# this
if (len(dimensions) == 2):	
    dimensions = (dimensions[1], dimensions[0])	
else:	
    dimensions = (dimensions[2], dimensions[1], dimensions[0])

# is the same as this:
dimensions = tuple(dimensions[::-1])
```
that's using the start:stop:step syntax of the item getter: i.e. "go from the beginning to the end in steps of -1"

haesleinhuepf:
> that's using the start:stop:step syntax of the item getter: i.e. "go from the beginning to the end in steps of -1"

I will never get used to that 🤣

haesleinhuepf:
Awesome, big thanks for helping me cleaning my mess 🙂

PR #22, opened 2020-07-16
Add CI testing with github actions

tlambert03:
This PR will add CI tests with github actions.  In past projects, it's been a little tricky to get openCL working on CI, cross-platform... but this method has worked for me before on mac & linux... but still working on windows (I may just skip windows CI for now if too hard)

I've marked all of the currently failing tests with [`pytest.mark.xfail`](https://docs.pytest.org/en/latest/reference.html#pytest-mark-xfail-ref), so we can at least get some tests through.  In future PRs, we can address each of those failures separately.  One class of fails comes from pyopencl `BUILD_PROGRAM_FAILURE`s ... whereas the other ('cl.LogicError'/'INVALID_ARG_SIZE')only happen on Linux CI, presumably due to some limitations of hardware and/or drivers.  In the past, it's been a little tricky to get everything working on CI that also works locally... but, baby steps :)


tlambert03:
note: until this merges, and the .github folder is in your repo, this will not trigger any CI.  So you'll have to look at my own repo to watch tests for now:
https://github.com/tlambert-forks/pyclesperanto_prototype/actions

tlambert03:
I'm getting `PLATFORM_NOT_FOUND_KHR`  errors on github actions with windows.   The pyopencl package itself uses github actions, and doesn't test windows ...  so I think it's probably best to tackle that later.  This is all set for review.

I do also want to point out one important fix:
https://github.com/clEsperanto/pyclesperanto_prototype/pull/22/files#diff-94e3ef33f8614e88ecf2f7574a4a3046
in which a difference between the gputools `get_device` function and the new built-in one was breaking the OCLProgram

tlambert03:
This also includes some fixes to the setup.py and pip install process that were introduced in the reorganization

haesleinhuepf:
> I'm getting `PLATFORM_NOT_FOUND_KHR` errors on github actions with windows. The pyopencl package itself uses github actions,

I'm having issues with pyopencl on Windows since the very beginning. 

I will just merge everything and see how stable it is now :-)

Thanks again for all this effort! I would be lost with CI...

PR #16, opened 2020-07-15
Add tests

tlambert03:
closes #6  
starting work on tests with pytest.  To try them out, pull this PR, do `pip install pytest pytest-benchmark`, then run `python -m pytest`.  (after there is a setup.py file in here, you can install the package locally with `pip install -e .` and then just "pytest" alone will work).

I added an example benchmark to compare CPU with GPU at different array sizes for test_multiply_matrix ... which do look like there's some work to be done  😅
```
---------------------------------------- benchmark: 16 tests --------------------------
Name (time in us)                              Mean                 StdDev       Rounds
---------------------------------------------------------------------------------------
test_multiply_matrix[cpu-64x64]             70.9399 (1.0)           7.6045         6413
test_multiply_matrix[gpu-64x64]          3,693.2913 (52.06)       259.3121          230
test_multiply_matrix[cpu-128x128]          267.7748 (3.77)         26.8374         2153
test_multiply_matrix[gpu-128x128]        3,987.1152 (56.20)       270.0672          240
test_multiply_matrix[cpu-256x256]        1,139.2580 (16.06)       180.4972          619
test_multiply_matrix[gpu-256x256]        5,276.0616 (74.37)       412.6583          164
test_multiply_matrix[cpu-512x512]        4,861.8999 (68.54)       601.1902          158
test_multiply_matrix[gpu-512x512]       11,101.6707 (156.49)      235.0929           76
test_multiply_matrix[cpu-1024x1024]     21,199.7104 (298.84)    1,290.4645           30
test_multiply_matrix[gpu-1024x1024]     44,066.2162 (621.18)    1,059.5423           20
test_multiply_matrix[cpu-2048x2048]    126,025.3478 (>1000.0)   8,929.1151            6
test_multiply_matrix[gpu-2048x2048]    210,418.4442 (>1000.0)  11,606.9894            5
test_multiply_matrix[cpu-4096x4096]    814,364.7442 (>1000.0)  42,047.9409            5
test_multiply_matrix[gpu-4096x4096]  1,161,997.9468 (>1000.0)  16,589.3983            5
test_multiply_matrix[cpu-8192x8192]  4,916,867.3818 (>1000.0)  56,979.3099            5
test_multiply_matrix[gpu-8192x8192]  6,693,529.5528 (>1000.0)  65,086.0641            5
----------------------------------------------------------------------------------------
```

`test_kernels` still needs a lot of work, and feel free to directly edit this PR if you're so inclined.

PR #23, opened 2020-07-17
3D segmentation workflow

haesleinhuepf:
This brings everything one needs for the [famous tribolium-workflow](https://github.com/clEsperanto/pyclesperanto_prototype/blob/cca/demo/tribolium_morphometry/Tribolium%20morphometry.ipynb). I'm happy that the [plugin_function](https://github.com/clEsperanto/pyclesperanto_prototype/blob/cca/pyclesperanto_prototype/_tier0/plugin_function.py) (suggested / inspired by @jni ) decorator works well. I'm going to introduce it everywhere next.

Performance is horrible. We clearly have an issue here. :-)

This PR also brings tons for bugfixes and is quite bulky. If no one sees some dramatic isses, I will just merge soon.


tlambert03:
before you go too far fixing deployment, maybe look at/merge #16, which also fixes them

tlambert03:
regarding package data.... just use `include_package_data=True` and leave it to MANIFEST.in to declare them

tlambert03:
I was playing with caching locally as well.  I agree it's worth considering, but I will point out that the program building step usually takes about 7 microseconds (on my computer), so while gputools did cache those programs, it's only going to save on the order of microseconds. 

second is that if you do want to cache them, we might be able to do it slightly cleaner using [`@lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache), which would both calculate & cache a program given a set of kernel name, parameters, etc...

I'll follow up on that later

haesleinhuepf:
I was just implementing the caching strategy from clij. Happy to improve on this. I also remember that we had the suspicion that gputools-caching wouldn't work here, but I can't recall why... It had something to do with the keys in the cache....

haesleinhuepf:
> second is that if you do want to cache them, we might be able to do it slightly cleaner using [`@lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache), which would both calculate & cache a program given a set of kernel name, parameters, etc...

Ok. 😳  This hint is GOLD! I took weeks for implementing this on the Java side. Here it's built-in 🤯 
🙏 @tlambert03

Issue #10, opened 2020-06-04
Function-based interface versus object-oriented approach


haesleinhuepf:
Dear clesperanto-fans, dear future self,

at the moment, we're using [static functions](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/clesperanto/plugins/__init__.py) to call operations. The GPU device selection is global and transparent. In a normal user-scenario, we don't event know what GPU is selected by gputools. In CLIJ2, all operations sit in a [code-generated interface](https://github.com/clij/clij2/blob/master/src/main/java/net/haesleinhuepf/clij2/CLIJ2Ops.java) that's attached to an CLIJ2-instance which allows us to have several instances executing operations in parallel on several GPUs. We may have to switch to a similar approach in clEsperanto.

Cheers,
Robert

jni:
Why are instances required as opposed to eg a `_device=` keyword argument?

haesleinhuepf:
Both would work. Actually, both is implemented in [CLIJ2](https://github.com/clij/clij2/blob/master/src/main/java/net/haesleinhuepf/clij2/plugins/AddImages.java#L27). All operations exist as static methods can can be called with a CLIJ(2) instance. I'm not advertising this as public API for two reasons: Auto-completion allows you to browse available methods if you start from the clij2.gateway. You know discovering what operations are there is the major problem for both beginners and experts ;-)
![image](https://user-images.githubusercontent.com/12660498/83752022-1c4f7380-a668-11ea-843e-107df450f0c0.png)

Furthermore, it's intuitive to call and see that everyting happens in "clij2":
```
a = clij2.push()
clij2.process(a, b)
clij2.pull(b)
```

Background is, you cannot process images on one device which were pushed to another. This would crash in the last line:
```
clijA = CLIJ2("rtx")
clijB = CLIJ2("gtx")

imageA = clijA.push(image_cpu)
imageB = clijB.create(imageA)

clijA.absolute(imageA, imageB)
```

Btw. in general, I would love to dynamically distribute tasks between GPUs, but various approaches failed... It boils down to the end-user who makes the decision on which device to do what.

haesleinhuepf:
To make this clear: I would be happy to have both approaches available to end-users. They should decide what they like most. I actually love to do code like this:
```
import cleasperanto as cle
cle.operation(...)
```

It's very beautiful and simple in contrast to
```
import clesperanto
cle = clesperanto.getInstance()

cle.operation()
```

[...]

PR #27, opened 2020-07-18
Cleanup get_device() and add select_device()

### LINKED ISSUE #26: Enable user to do GPU-selection
[get_gpu](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/pyclesperanto_prototype/_tier0/_pycl.py#L20-L22) per default selects the GPU with most memory. On a system with an NVidia GPU (11 GB memory) and 64 GB RAM, it chooses Intels GPU. 

In CLIJ, the use has the [optional choice](https://github.com/clij/clij-core/blob/master/src/main/java/net/haesleinhuepf/clij/CLIJ.java#L170) to enter a device name or index.

tlambert03:
closes #26 
This PR does a couple things around device selection:
- standardize on usage of `cle.get_device` (instead of `get_gpu` since the device may be a CPU)
- create `Device` container on which we can put methods like `__repr__` and `name`.  So rather than creating a new `get_device_name` function, we just use `get_device().name` ... also just doing `get_device()` on it's own will print a nice representation now, like `<GeForce GTX 1080 Ti on Platform: Apple (2 refs)>`
- add new `select_device` function that takes three optional parameters `name` (first device containing the string will be returned), `dev_type` (`gpu` or `cpu`), and a `score_key` function that takes a `cl.Device` object and returns a float, so users can control how their devices will be sorted.  These can each be used alone or together

I also put the program caching as a classmethod on the `OCLPRogram` class, so now you just use `OCLProgram.from_source(src)`.  And I reverted the changes made in #23 that stored the context and queue on the device (that issue was solved elsewhere in the code in #22).


tlambert03:
another thought:  I noticed a bunch of new things that were pushed directly to master, and/or included in big PRs that seemed unrelated to a specific group of changes.   It can sometimes be a bit hard to comment/contribute/collaborate in those cases.   It's still a young project though, and it may be that you'd prefer to iterate faster on this than is allowed with the typical github PR/review pattern (and that would be totally fine! it's your project!! 😄 ) but just wanted to mention that.

haesleinhuepf:
It's good that you say that, because I'm new to this working style. I'm not so much a framework/platform developer. I rather implement use-cases, which often involves changes on multiple levels. Then, after some saturation (many use cases run smoothly), I fill the gaps. That's unusual I guess? 😁
Would it make sense to split these things into multiple PRs with specific topics?

tlambert03:
not unusual at all! and I'm not sure there's a fundamental difference between what we do when we're working on specific use-cases for users vs designing packages.  There's plenty of top-down design in packages as well, and often we need to touch large parts of the package for a given change.  I'd say it's more about collaborative coding vs working alone... 

I'd _also_ say that it can be rather _annoying_ to have to think about others looking at the code 😆 , particularly in the early stages of a project (where it can slow you down to have to organize changes into modular chunks), which is why I wouldn't fault you one bit if you want to keep going for now with your current workflow, until things settle a bit more.  I'll keep checking in either case

> Would it make sense to split these things into multiple PRs with specific topics?

yeah, more or less.  for instance when working on the tribolium example, try not to change how the program is cached, or the characteristics of fundamental objects like the device.  And as you work, if you come across something that should be changed, but perhaps isn't critical to be changed for the current feature you're working on, create a quick branch and start the change... then come back to your feature, and push them as two separate PRs.  (again, acknowledging that that can be annoying and totally understand if you don't want to!). 
also, if possible, don't work off of or push to master... that becomes important if two people are trying to debug something and you can have some common working reference point (where tests are passing).

tlambert03:
@jni has tons of experience with this and may have additional thoughts on how best to strike a good balance between the need for rapid iteration in what starts out as a personal repo with the need for modularity and review in collaborative coding.

haesleinhuepf:
> And as you work, if you come across something that should be changed, but perhaps isn't critical to be changed for the current feature you're working on, create a quick branch and start the change... then come back to your feature, and push them as two separate PRs. (again, acknowledging that that can be annoying and totally understand if you don't want to!).

Ok, I have to practice that. Thanks for your patience and support! Really appreciated! 

> * create `Device` container on which we can put methods like `__repr__` and `name`.

That sounds very interesting. Long term idea: Can we optionally pass this `Device` to the execute method? That would allow us to run on multiple GPUs at a time. There were voices saying that this isn't possible, but I would love to give this a try. What would you say? Have you ever tried pyopencl from multiple threads?

tlambert03:
> Thanks for your patience and support! Really appreciated!

you bet. I have only recently/gradually picked some of this up myself by annoying @jni enough times by merging too quickly on side projects 😂 .  It's definitely a whole different model and it does come with its downsides.  (though I think they are far outweighed by the upsides)

> Long term idea: Can we optionally pass this `Device` to the execute method?

yeah, I actually put an "open question" [comment in this PR](https://github.com/clEsperanto/pyclesperanto_prototype/pull/27/files#diff-94e3ef33f8614e88ecf2f7574a4a3046R11-R12) about what we should/shouldn't be passing around.  I think a deep dive into the `pyopencl` docs might be revealing (and I kinda wonder whether it's the [Context](https://documen.tician.de/pyopencl/runtime_platform.html#context) we should be passing around more than the [Device](https://documen.tician.de/pyopencl/runtime_platform.html#device) object,  since that's what we're really extracting out of the Device object when we create a [Program](https://documen.tician.de/pyopencl/runtime_program.html)).  There are also a lot of convenience methods on those classes that we might not be taking full advantage of 

> That would allow us to run on multiple GPUs at a time. There were voices saying that this isn't possible, but I would love to give this a try. What would you say? Have you ever tried pyopencl from multiple threads?

I have not done that with pyopencl, but I have done multi-GPU stuff with cuda in [LLSpy](https://github.com/tlambert03/LLSpy). (sidenote: I do _not_ ever recommend looking at the source code of LLSpy for patterns 😂 ), and there it wasn't multithreading, but rather multi-processing.  The use case there lended itself well to spreading work across GPUs, as the work itself was independent and didn't need any shared host memory, etc... So I don't have any particular insight into how hard this will be here.  it will likely depend on the exact parallelization strategy.

haesleinhuepf:
> yeah, I actually put an "open question" [comment in this PR](https://github.com/clEsperanto/pyclesperanto_prototype/pull/27/files#diff-94e3ef33f8614e88ecf2f7574a4a3046R11-R12) about what we should/shouldn't be passing around.

That fits in an [earlier discussion](https://github.com/clEsperanto/pyclesperanto_prototype/issues/10)

haesleinhuepf:
> it will likely depend on the exact parallelization strategy.

Yes, it's super tricky. I spent lot of time on that. What works quite ok: distributing timepoints/channels of light sheet data to multiple GPUs and do the same operations on them. It appears to me that there is no general / transparent solution... Side note: It's also [evil in some cases](https://github.com/clij/clijx/commit/b251d133a033375e4e79514fbc915f1dc343102b#diff-6d398e6d01b0ae2ae47732aece936092R17-R18)...

tlambert03:
ready for review

haesleinhuepf:
Ok, this looks great! I just tried to run tests locally and there is an issue (which I thought I saw ealier and was fixed elsewhere)...

Do you know what that means?

```
C:\structure\code\pyclesperanto_prototype\tests\test_multiply_matrix.py:6
__________________________________________ ERROR at setup of test_multiply_matrix[cpu-1024x1024] ___________________________________________
file C:\structure\code\pyclesperanto_prototype\tests\test_multiply_matrix.py, line 6
  @pytest.mark.parametrize("shape", 2 ** np.arange(9, 11), ids=lambda x: f"{x}x{x}")
E       fixture 'benchmark' not found
>       available fixtures: cache, capfd, capfdbinary, caplog, capsys, capsysbinary, doctest_namespace, monkeypatch, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.
```

What is a "fixture"?

I'm happy to merge it anyway. This appears to be just a local problem on my machine...

haesleinhuepf:
> What is a "fixture"?

I mean, I can [found this](https://docs.pytest.org/en/stable/fixture.html). But not sure how to get this test work here... Do I have to specify pytest-benchmark as a dependency in test-scope somewhere?

tlambert03:
lastly, it's typical in a python project to have `requirements.txt` files that provide additional dependencies (outside of your package runtime dependencies) for running tests, building docs, etc... so we should probably just add a `requirements.txt` file that gets a totally clean environment ready to go for development

haesleinhuepf:
> that probably means you need to install pytest-benchmark `pip install pytest-benchmark` (see instructions from #16)

Here you see that I'm switching between computers all the time. 

> lastly, it's typical in a python project to have `requirements.txt` files that provide additional dependencies (outside of your package runtime dependencies) for running tests, building docs, etc... so we should probably just add a `requirements.txt` file that gets a totally clean environment ready to go for development

I was about to ask about that :-) instead I'll create it right away.

tlambert03:
> Here you see that I'm switching between computers all the time.

yeah me too.  and I consider environments to be "disposable" ... so having a quick way to get fully installed is best.

> I'll create it right away.

Awesome!  one trick that might not be immediately obvious is to add `-e .` to the beginning of the file:

```text
-e .
pytest
pytest-benchmark
...
```
This will install the package in editable mode, pulling all of the standard requirements from your setup.py file, so you don't have to duplicate them in your requirements.txt file.

haesleinhuepf:
> one trick that might not be immediately obvious is to add `-e .`

Yeah. @jni taught me that trick today. I feel much less handicapped now :-)

PR #38, opened 2020-11-05
Make many tests pass in CI; remove xfails

haesleinhuepf:
I managed to make many tests pass in githubs CI. The reason was that the OpenCL-implementation (CPU based - [POCL](http://portablecl.org/)?) expects `int32` as integer parameters and not `int`. Thus, the actually crucial commit is [this one](https://github.com/clEsperanto/pyclesperanto_prototype/commit/9534674729e752d0ff05b708af6f2701205cdfe8). 

We will later also see if this also fixes pyclesperanto in google colab. If so, @oeway and @imjoy-team may be very happy ;-)

### DIFF
--- a/pyclesperanto_prototype/_tier0/_execute.py
+++ b/pyclesperanto_prototype/_tier0/_execute.py
@@ -115,7 +115,7 @@ def execute(anchor, opencl_kernel_filename, kernel_name, global_size, parameters
             defines.extend(IMAGE_HEADER.format(**params).split("\n"))
 
         elif isinstance(value, int):
-            arguments.append(np.array([value], np.int))
+            arguments.append(np.array([value], np.int32))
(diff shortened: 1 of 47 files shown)


haesleinhuepf:
Alright. @tlambert03 may like this as well. Thanks for setting up the awesome CI, Talley! I'm getting used to it 👍

jni:
> Alright. @tlambert03 may like this as well

Yes that was going to be my comment! :joy: Awesome detective work @haesleinhuepf! :tada:

tlambert03:
> I managed to make many tests pass in githubs CI. The reason was that the OpenCL-implementation (CPU based - [POCL](http://portablecl.org/)?) expects `int32` as integer parameters and not `int`. 

awesome!

tlambert03 (approved):

PR #42, opened 2020-11-18
List available opencl devices

haesleinhuepf:
This brings a convenience method for retrieving a list of names of available opencl devices as suggested by @tlambert03 . A [counter-part in clij](https://github.com/clij/clij-core/blob/master/src/main/java/net/haesleinhuepf/clij/CLIJ.java#L206) has the prefix "get" but @jni convinced me that we should get rid of all "get" prefixes as all methods return some sort of result.

### DIFF
--- a/pyclesperanto_prototype/_tier0/_available_device_names.py
+++ b/pyclesperanto_prototype/_tier0/_available_device_names.py
@@ -1,13 +1,30 @@
 from .._tier0._pycl import filter_devices
 from typing import List
 
-def available_device_names() -> List[str]:
+def available_device_names(*args, **kwargs) -> List[str]:
     """Retrieve a list of names of available OpenCL-devices
+
+    Arguments are forwarded to `filter_devices`.
 
     Returns
     -------
         list of OpenCL-device names
+
+    See Also
+    --------
+    filter_devices : Returns list of devices instead of device names
+
+    Examples
+    --------
+    >>> import pyclesperanto_prototype as cle
+    >>> gpu_devices = cle.available_device_names(dev_type="gpu")
+    >>> print("Available GPU OpenCL devices:" + str(gpu_devices))
+    >>>
+    >>> cpu_devices = cle.available_device_names(dev_type="cpu")
+    >>> print("Available CPU OpenCL devices:" + str(cpu_devices))
+
+
     """
-    devices = filter_devices()
+    devices = filter_devices(*args, **kwargs)
     device_names = [device.name for device in devices]
     return device_names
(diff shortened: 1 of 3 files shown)


jni (approved):

jni:
Bleurgh. CI is broken by some GitHub Actions deprecations. See this line:

https://github.com/clEsperanto/pyclesperanto_prototype/pull/42/checks?check_run_id=1417118665#step:3:19

and this blog post:

https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

I don't have time right now to investigate a fix, but hopefully I've saved you a click or two! 😂

tlambert03 (commented):
looks good to

pyclesperanto_prototype/_tier0/_available_device_names.py
```diff
@@ -0,0 +1,13 @@
+from .._tier0._pycl import filter_devices
+from typing import List
+
+def available_device_names() -> List[str]:
```
tlambert03, line 4: maybe pass `**kwargs` to filter_devices (and indicate that it is being done in the docs) so that someone could say `available_device_names(dev_type="gpu")` for instance

tlambert03 (commented):
looks good to me, with one minor suggestion about passing kwargs

tlambert03 (approved):

haesleinhuepf:
Great idea @tlambert03. I'm wondering if a "see also" section is enough to enable easy exploration of what parameters are available or if I should specify parameters. This would mean duplication of documentation and I'm just not sure if that makes sense.

tlambert03:
> Great idea @tlambert03. I'm wondering if a "see also" section is enough to enable easy exploration of what parameters are available or if I should specify parameters. This would mean duplication of documentation and I'm just not sure if that makes sense.

yeah... the eternal question :joy:
I'm not sure there's a "right" answer here, duplicating docs makes it easier for someone in interactive usage to the info quickly, but obviously increases the burden on developers to prevent future inconsistency.  In this case, since `filter_devices` is so hard to get to (in that `_tier0` module), I might _slightly_ favor duplication... but whatever you prefer is fine with me.
(sidenote: I want to get back to the discussion of package structure someday! :) I love the tier system for clarifying cl kernel dependencies ... but not as much for the core python code that ties it all together)

haesleinhuepf:
Ok, I basically copied over parameters and comments. I left out the name 'name' parameter because if specified it would only return the first device with that name, which doesn't fit to what the method is supposed to do. We may refactor those selection methods later. For now now I'm happy with the functionality.

Issue #33, opened 2020-07-24
build a numpy-like interface


yxdragon:
hi, when I push array to gpu, got a gpu object.
can we wrap the gpu object as a numpy-like interface? such as (dot, concat, slice, reshape...)?
I know there is a repo called clpy, but it is too hard to install.

haesleinhuepf:
Hey @yxdragon ,

that's a great idea! As we have it on [matlab](https://github.com/clij/clatlab/blob/master/src/main/java/net/haesleinhuepf/clatlab/MOCLBuffer.java) side, it totally makes sense here as well :-)

Thanks for the suggestion! If you feel like spending some time on it, contributions are very welcome!

Cheers,
Robert

haesleinhuepf:
Just adding a link here for how this could be implemented (thanks to @jni for pointing me towards this):
https://data-apis.github.io/array-api/latest/API_specification/array_object.html

PR #48, opened 2020-11-25
added example using magicgui

haesleinhuepf:
This just adds a magicgui demo for using pyclesperanto in napari

@tlambert03 is there a way to make magicgui add a labels-layer instead of an image layer?

Thanks!

### DIFF
--- a/demo/napari_gui/napari_.py
+++ b/demo/napari_gui/napari_.py
@@ -1,0 +1,26 @@
+# load data
+from skimage.io import imread
+image = imread('https://samples.fiji.sc/blobs.png')
+
+# start up napari
+import napari
+import pyclesperanto_prototype as cle
+with napari.gui_qt():
+    viewer = napari.Viewer()
+    viewer.add_image(image, name='blobs')
+
+    # push image to GPU
+    input = cle.push_zyx(image)
+
+    # process the mage
+    sigma = 3
+    blurred = cle.gaussian_blur(input, sigma_x=sigma, sigma_y=sigma)
+    binary = cle.threshold_otsu(blurred)
+    labels = cle.connected_components_labeling_box(binary)
+
+    # pull result back
+    output = cle.pull_zyx(labels)
+
+    # add it to napari
+    viewer.add_labels(output)
+
--- a/demo/napari_gui/napari_.py
+++ b/demo/napari_gui/napari_.py
@@ -1,10 +1,11 @@
+import pyclesperanto_prototype as cle
+
 # load data
 from skimage.io import imread
 image = imread('https://samples.fiji.sc/blobs.png')
 
 # start up napari
 import napari
-import pyclesperanto_prototype as cle
 with napari.gui_qt():
     viewer = napari.Viewer()
     viewer.add_image(image, name='blobs')
@@ -12,7 +13,7 @@ with napari.gui_qt():
     # push image to GPU
     input = cle.push_zyx(image)
 
-    # process the mage
+    # process the image
     sigma = 3
     blurred = cle.gaussian_blur(input, sigma_x=sigma, sigma_y=sigma)
     binary = cle.threshold_otsu(blurred)
--- a/demo/napari_gui/napari_magicgui.py
+++ b/demo/napari_gui/napari_magicgui.py
@@ -11,7 +11,7 @@ def process_image(input: Image, sigma: float = 5) -> Labels:
         # push image to GPU
         input = cle.push_zyx(input.data)
 
-        # process the mage
+        # process the image
         blurred = cle.gaussian_blur(input, sigma_x=sigma, sigma_y=sigma)
         binary = cle.threshold_otsu(blurred)
         labels = cle.connected_components_labeling_box(binary)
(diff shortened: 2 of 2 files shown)


tlambert03:
Yes there is, if just changing the return annotation to Label doesn’t do it, there’s a more thorough way that I can show you (but can’t right now)

haesleinhuepf:
Awesome @tlambert03  that was it. I tried changing the annotation _and_ the returned object. But obviously, only changing the annotation alone does the job. It's magic. 

Thanks!

demo/napari_gui/napari_.py
```diff
@@ -0,0 +1,26 @@
...
+import napari
+import pyclesperanto_prototype as cle
+with napari.gui_qt():
+    viewer = napari.Viewer()
+    viewer.add_image(image, name='blobs')
+
+    # push image to GPU
+    input = cle.push_zyx(image)
+
+    # process the mage
```
jni, line 15: 🏭🧙

demo/napari_gui/napari_.py
```diff
@@ -0,0 +1,26 @@
...
+    sigma = 3
+    blurred = cle.gaussian_blur(input, sigma_x=sigma, sigma_y=sigma)
+    binary = cle.threshold_otsu(blurred)
+    labels = cle.connected_components_labeling_box(binary)
+
+    # pull result back
+    output = cle.pull_zyx(labels)
```
jni, line 22: Can this step be skipped because of `__array__`?

demo/napari_gui/napari_.py
```diff
@@ -0,0 +1,26 @@
...
+with napari.gui_qt():
+    viewer = napari.Viewer()
+    viewer.add_image(image, name='blobs')
+
+    # push image to GPU
+    input = cle.push_zyx(image)
+
+    # process the mage
+    sigma = 3
+    blurred = cle.gaussian_blur(input, sigma_x=sigma, sigma_y=sigma)
```
jni, line 17: `sigma_x` makes me cringe. What do you think about refining the API here to allow vector sigma corresponding to the array axes?

demo/napari_gui/napari_.py
```diff
@@ -0,0 +1,26 @@
...
+    binary = cle.threshold_otsu(blurred)
+    labels = cle.connected_components_labeling_box(binary)
```
jni, line 19: Why box? 😂

demo/napari_gui/napari_magicgui.py
```diff
@@ -0,0 +1,36 @@
...
+from magicgui import magicgui
+from napari.layers import Image, Labels
+
+@magicgui(auto_call=True)
+def process_image(input: Image, sigma: float = 5) -> Labels:
+    if input:
+        # push image to GPU
+        input = cle.push_zyx(input.data)
+
+        # process the mage
```
jni, line 14: 🏭🧙

demo/napari_gui/napari_magicgui.py
```diff
@@ -0,0 +1,36 @@
...
+import napari
+import pyclesperanto_prototype as cle
+from magicgui import magicgui
+from napari.layers import Image, Labels
+
+@magicgui(auto_call=True)
+def process_image(input: Image, sigma: float = 5) -> Labels:
+    if input:
+        # push image to GPU
+        input = cle.push_zyx(input.data)
```
jni, line 12: Does `cle` autopush? And if so, can we take advantage of that here? ;) And if not, can we raise an issue? =P

jni (approved):
Very exciting stuff. =) I've left some comments @haesleinhuepf!

demo/napari_gui/napari_.py
haesleinhuepf: likely yes. I'd like to keep it for the moment to be consistent with the other languages/dialects. But I strongly agree with @royerloic - it would be awesome to make clesperanto more transparent and numpy-like. Push and pull might fall first when implementing this.
See:
https://twitter.com/loicaroyer/status/1331712368024272897?s=20
and
https://github.com/clEsperanto/pyclesperanto_prototype/issues/33

demo/napari_gui/napari_.py
haesleinhuepf: :P 
Box: 8- resp. 26-connected (3d/2d), partly separable
Diamond: 4- resp. 6-connected, not separable

I engineered this terminology very carefully. And I'd love to know how to improve it.

demo/napari_gui/napari_magicgui.py
haesleinhuepf: it auto-pushes, but it doesn't do push_zyx. I'll create and issue on this: https://github.com/clEsperanto/pyclesperanto_prototype/issues/49

demo/napari_gui/napari_.py
jni: I prefer to use `connected_components_labeling(binary, connectivity={1...ndim})` (see [`ndimage.label`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.label.html)/[`ndimage.generate_binary_structure`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.generate_binary_structure.html#scipy.ndimage.generate_binary_structure)/[`skimage.measure.label`](https://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.label)). Separability is an implementation detail that you can handle internally (when connectivity==ndim).

PR #51, opened 2020-11-28
Sync with clij2

haesleinhuepf:
This is an overdued PR resulting from asychronous development of CLIJ2 and pyclesperanto. A lot of kernels were out of sync. This should not change any functionality, just fix bugs discovered in CLIJ2 and improve naming conventions of cl files.


tlambert03:
just curious: the ultimate goal here would be to have a single repo of `.cl` kernels right, used by all of the various language interfaces, yeah?  What are the current barriers to that?

haesleinhuepf:
> just curious: the ultimate goal here would be to have a single repo of `.cl` kernels right, used by all of the various language interfaces, yeah?

Yes, absolutely! 

> What are the current barriers to that?

My inability to deal with git-submodules. @StRigaud implemented that very nicely in the [C++ counterpart](https://github.com/clEsperanto/CLIc_prototype/blob/master/.gitmodules) and I would strive for reusing the same strategy. Therefore, it might make sense to remove the java-folder hierarchy from the [clij-opencl-kernels](https://github.com/clEsperanto/clij-opencl-kernels/tree/master/src/main/java/net/haesleinhuepf/clij/kernels).

However, there is an alternate strategy to discuss: We may want to have a folder for each cl file (a "module") which contains also a text file with documentation and potentially wrapper files, e.g. in python, Java or C++. @StRigaud and myself were thinking of diving a bit deeper into ITK to see how they managed similar stuff. 

I assume that's something we do at some point after I2K ;-)

tlambert03:
> git-submodules... Therefore, it might make sense to remove the java-folder hierarchy from the clij-opencl-kernels.

awesome.  yeah, I'd love to see all of those `.cl` files pulled out of the java repo, and have clij also use submodules to access the kernels

> We may want to have a folder for each cl file (a "module") which contains also a text file with documentation and potentially wrapper files, e.g. in python, Java or C++.

as I think you know, I personally won't be super happy on the python side until all of that is autogenerated from the kernels (which should include the docs and stuff).  I _don't_ want to see individual `.py` files for every `.cl` file ... and I maintain that it shouldn't be necessary (in python at least ... not sure about the others).  But curious to hear what you and @StRigaud learn after diving in.  I'll let you know if/when I get around to prototyping a python strategy for autogeneration

PR #64, opened 2020-12-25
Measure intensity variance and standard deviation of labels and store in regionprops

haesleinhuepf:
Hi all,

I'm aware that it's holidays. No hurry.

This PR allows us to measure the standard deviation intensity of labelled regions. It's a bit expensive computationally  as we need to call skimage.measure.regionprops twice. On Java side we have similar issues.

This PR may become obsolete depending on answers in the related discussion here:
https://forum.image.sc/t/how-to-measure-standard-deviation-of-intensities-with-scikit-image-regionprops/46948

Merry Christmas!

Cheers,
Robert


tests/test_statistics_of_labelled_pixels.py
```diff
@@ -218,6 +218,28 @@ def test_statistics_of_labelled_pixels_3d():
...
+    ])
+
+    print("numpy standard deviation", np.std(image))
+    print("numpy variance", np.var(image))
+
+    stats = cle.statistics_of_labelled_pixels(image, labels)
+
+    print("std", stats[0].standard_deviation_intensity)
+
+    assert np.abs(stats[0].standard_deviation_intensity - 2.58) < 0.1
```
jni, line 243: You can do `from numpy import testing` then `testing.assert_almost_equal(stats[0].standard_deviation_intensity, 2.58, decimal=1)`

jni: (which has nice features like telling you the value and deviation when the test fails)

pyclesperanto_prototype/_tier9/_statistics_of_background_and_labelled_pixels.py
```diff
@@ -6,7 +6,7 @@
 from .._tier0 import pull_zyx
 
 @plugin_function(output_creator=create_none)
-def statistics_of_background_and_labelled_pixels(input : Image, labelmap : Image):
+def statistics_of_background_and_labelled_pixels(input : Image, labelmap : Image, extra_properties=[]):
```
jni, line 9: This is fine although note that skimage allows measuring regionprops *without* pixel intensities, and this is useful. Not sure whether you want to support that use case...

jni: also fwiw the American spelling is `labeled`. I don't have a strong opinion on US vs UK spelling but Clesperanto should pick one and stick to it. Generally I see most people using US in code. (e.g. neighbors vs neighbours).

tests/test_statistics_of_labelled_pixels.py
haesleinhuepf: Awesome! I continure learning, which is great :-)

pyclesperanto_prototype/_tier9/_statistics_of_background_and_labelled_pixels.py
haesleinhuepf: I see, because it's connected_components_labeling with one L. Pitty. I created as issue for this #66. I assume there are many more methods that could have better names (e.g. more similar to scikit-image, or imagej or matlab), so we can collect some ideas like this one. Thanks again!

haesleinhuepf: I have an idea: How about German method names? Instead of "connected_component_labeling" it would then be "kennzeichnung_verbundener_komponenten". That might reduce typos and mistakes on my side dramatically 🤣

PR #74, opened 2021-01-01
rename flag_existing_intensities to flag_existing_labels

haesleinhuepf:
I'm renaming the method `flag_existing_intensities' to `flag_existing_labels`, because that explains better what the method can be used for. I also created an alias to be backwards-compatible.


haesleinhuepf:
Note to self: This line doesn't need to be. The same is done inside flag_existing_intensities.
https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/pyclesperanto_prototype/_tier3/_close_index_gaps_in_label_map.py#L42

PR #75, opened 2021-01-02
Use git submodule for kernels

haesleinhuepf:
This PR deletes all .cl files from the project and uses a git submodule of the clij-opencl-kernels repository instead. @tlambert03 and @StRigaud  may be happy to hear that :-)


tlambert03:
awesome!!

PR #79, opened 2021-01-03
Speedup measurements (statistics_of_labelled_pixels a.k.a. regionprops)

haesleinhuepf:
This brings a GPU version of statistics_of_labelled_pixels. It does not use scikit-image regionprops anymore and also has a different output format. It's a dictionary now, comparable to regionprops_table.

It is still possible to use the old implementation by setting `use_gpu=False`. There are warnings mentioning that the old implementation is deprecated and will be removed in a future version.

I also added a simple [benchmarking notebook](https://github.com/clEsperanto/pyclesperanto_prototype/blob/speedup_measurements/benchmarks/benchmark_statistics_of_labeled_pixels.ipynb) to demonstrate how much time one can spare. Especially in 3D it makes sense.


pyclesperanto_prototype/_tier9/_statistics_of_image.py
```diff
@@ -2,7 +2,7 @@
 from .._tier0 import plugin_function
 
 @plugin_function
-def statistics_of_image(image : Image):
+def statistics_of_image(image : Image, use_gpu:bool = False):
```
jni, line 5: Note, for arguments like this one, I would recommend `(image : Image, *, use_gpu : bool = False)`, which forces users to use keyword arguments, `statistics_of_image(my_image, use_gpu=True)`. This has two advantages:

- user code is more self-documenting. Compare the above to `statistics_of_image(my_image, True)` — what does `True` mean here???
- if you ever want to deprecate the use of that argument, it becomes much easier to track whether user code intended to pass that argument or some future positional argument, e.g. `statistics_of_image(image : Image, mask : Image = None, *, use_gpu : bool = False)` becomes easier to get to without breaking user code.

pyclesperanto_prototype/_tier9/_statistics_of_labelled_pixels.py
```diff
@@ -1,32 +1,46 @@
...
+
 from skimage.measure import regionprops
 
 from .._tier0 import Image
 from .._tier0 import create_none
 from .._tier0 import plugin_function
 
 @plugin_function(output_creator=create_none)
-def statistics_of_labelled_pixels(input : Image = None, labelmap : Image = None, measure_shape : bool = True, extra_properties = []):
+def statistics_of_labelled_pixels(input : Image = None, labelmap : Image = None, measure_shape : bool = True, extra_properties = [], use_gpu : bool = True):
```
jni, line 10: See above. Note also that in skimage we allow `input` to be None, since many region properties can be measured without the image intensities. Also note that `input()` is a Python built-in function, so it's best to avoid its use as a keyword argument. (Yes, `scipy.ndimage` uses it, but this has led to multiple debates over the years about deprecating it.)

pyclesperanto_prototype/_tier9/_statistics_of_labelled_pixels.py
```diff
@@ -92,3 +106,284 @@ def standard_deviation_intensity(region, intensities):
 
     return props
 
+def _statistics_of_labelled_pixels_gpu(intensity_image : Image = None, label_image : Image = None, measure_shape : bool = True):
+    """
+    This method will replate statistics_of_labelled_pixels(use_gpu=True) at some point.
```
jni, line 111:
```suggestion
    This method will replace statistics_of_labelled_pixels(use_gpu=True) at some point.
```

pyclesperanto_prototype/_tier9/_statistics_of_labelled_pixels.py
```diff
@@ -92,3 +106,284 @@ def standard_deviation_intensity(region, intensities):
...
+    min_value = np.finfo(np.float32).min
+    max_value = np.finfo(np.float32).max
+    set_plane(sum_per_label_image, 8, max_value)
+    set_plane(sum_per_label_image, 9, min_value)
+    set_plane(sum_per_label_image, 10, max_value)
+    set_plane(sum_per_label_image, 11, min_value)
+    set_plane(sum_per_label_image, 12, max_value)
+    set_plane(sum_per_label_image, 13, min_value)
+    set_plane(sum_per_label_image, 14, max_value)
+    set_plane(sum_per_label_image, 15, min_value)
```
jni, line 177: Can you explain these lines? They are at the very least confusing! Not least because the planes 0-7 aren't touched for some reason?

pyclesperanto_prototype/_tier9/_statistics_of_labelled_pixels.py
```diff
@@ -92,3 +106,284 @@ def standard_deviation_intensity(region, intensities):
...
+
+    # accumulate statistics slice-by-slice
+    dimensions = [1, height, 1]
+    parameters = {
+        'dst': sum_per_label_image,
+        'src_label': label_image,
+        'src_image': intensity_image,
+        'sum_background': 0 # dpn't analyse background
```
jni, line 185:
```suggestion
        'sum_background': 0  # don't analyse background
```

pyclesperanto_prototype/_tier9/_statistics_of_labelled_pixels.py
haesleinhuepf: There is a comment two lines above explaining it. Furthermore, all lines (also 0-7) are touched, as they are set to 0. I'll rename the variable of that image, because the "sum" might be misleading.

https://github.com/clEsperanto/pyclesperanto_prototype/pull/79/files/4f6dd5a3e2f68344af6e788f15363e657cad9edc#diff-1c45f4e441422600bcfaa0213495cd24820ec27589879353d92de7b1ff911200R163-R167

pyclesperanto_prototype/_tier9/_statistics_of_image.py
haesleinhuepf: > which forces users to use keyword arguments

keyword arguments are not supported in the other clEsperanto dialects 😉 At least the mandatory arguments must be accessible via the minimal `operation(input1, input2)` syntax. I see what you mean. Maybe there is a way for improving tha API here..

jni: > must be accessible

I guess that's a problem for code generation, since exact uniformity is not possible? (?) But it should be possible to do something different for generated Python code vs generated [insert language I don't care about] code? 😜

PR #80, opened 2021-01-05
update scikit-image dep

tlambert03:
with the `extra_properties` argument used to `regionprops` in #64... I think we need to bump the scikit-image dependency to >=0.18.0...


haesleinhuepf:
Absolutely! Good catch @tlambert03 !

Issue #90, opened 2021-01-10
Affine transform: Shearing not implemented


haesleinhuepf:
In the current implementation of affine transforms, shearing has not been implemented. I don't use shearing that's why I'm not sure how the API should look like. In [CLIJ2 affine transforms](https://clij.github.io/clij2-docs/reference_affineTransform), we used that terminology:

shearXY=[factor]: shearing along X-axis in XY plane according to given factor
shearXZ=[factor]: shearing along X-axis in XZ plane according to given factor
shearYX=[factor]: shearing along Y-axis in XY plane according to given factor
shearYZ=[factor]: shearing along Y-axis in YZ plane according to given factor
shearZX=[factor]: shearing along Z-axis in XZ plane according to given factor
shearZY=[factor]: shearing along Z-axis in YZ plane according to given factor

I'm happy to implement this here again, but I'd need someone who tests it and provides feedback.

tlambert03:
Could you use an axes argument (tuple of 2 ints), as with done in the scipy.ndimage API? https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.rotate.html#scipy.ndimage.rotate

haesleinhuepf:
Hey @tlambert03 ,

this link points to rotation, not shearing. And I can't find shearing in scipy...

Anyway, I think I'll take a closer look on affine transforms in skimage. It might make sense to make the API compatible with this one.

haesleinhuepf:
... and I just realized that skimage only [supports 2D](https://scikit-image.org/docs/dev/api/skimage.transform.html#skimage.transform.AffineTransform) :-(

tlambert03:
Correct [`scipy.ndimage.interpolate`](https://docs.scipy.org/doc/scipy/reference/ndimage.html#interpolation) doesn't offer a `shear` convenience function (just `zoom`, `rotate`, and `shift` ... so I pointed to rotate as an example).  But the concept is the same: `axes` is a two-tuple of ints that defines the shear plane, (and ordering of provided axis matters to determine the shearing axis).  `factor` then would be a second argument that determines the magnitude.

... and yeah, skimage only supports 2D.

[...]

Issue #49, opened 2020-11-26
axis auto-transposition between python and opencl


haesleinhuepf:
Hi future-self, hi @jni , hi fans of axis-transposition,

coming back to your recent comments on https://github.com/clEsperanto/pyclesperanto_prototype/pull/48:

At the moment, I'm often using code like
```
image = cle.push_zyx(np_arr)

result = cle.gaussian_blur(image, sigma_x=3, sigma_y=4, sigma_z=5)
```

I'll do that because I'm heavily confused when it comes to multi-dimensional arrays. You know, in Java we use array[x][y][z], in CLIJ-OpenCL, we do it similarily and in matlab and python it's array[z][y][x]. The explicit terminology helps me debugging and maintaining functionality. This explicit x,y,z terminology appears not so popular in python code (subjective impression). 

I see we could change the API to:
```
result = cle.gaussian_blur(np_arr, [3, 4, 5])
```
or
```
result = cle.gaussian_blur(np_arr, [5, 4, 3])
```
(I'm really not sure!)

I'm just afraid that changing this would cause weird behaviour when people switch between eco-systems (like I do all the time). Thus, I'm wondering how to make things clear and ideally self-explanatory. Maybe even backwards-compatible.

Every best-practice, hint, suggestion is welcome. [How] do other libraries deal with such issues?

Cheers,
Robert

jni:
@haesleinhuepf I agree it's a mess and difficult, but for sure parameter names should *not* have any kind of implicit order that is not also encoded in the array. Specifically, the naming of axes is not present in any way in the NumPy array, or in the PyOpenCL Image either, I think. The only thing that we *do* know is the axis order, and so by using a list for the sigmas, which is itself ordered, we can keep track of which sigma corresponds to which axis. By contrast, using `sigma_x=`, I literally have *no idea* which axis you are talking about.

The clearest exposition of the arbitrariness of axis ordering I've come across is this one:

https://github.com/scikit-image/scikit-image/pull/3031#issuecomment-390941115

(You can scroll above it for more context, but the context is very long 😂.)

haesleinhuepf:
I see. And I would love to agree. My only concern is that we speak so much about z-planes, x/y/z coordinates, rows, columns, image-width and -height. Thus, there should be something corresponding in the API. 
I'll see if I can make an optional/additional `sigma` parameter, wherever there are `sigma_x`, `sigma_y` and `sigma_z`. I think it's doable with some python magic 😉

[...]

PR #91, opened 2021-01-10
Add affine transforms

haesleinhuepf:
Hey Talley @tlambert03 and Juan @jni ,

this brings affine transforms to pyclesperanto + some convenience functions for scaling, rotation, translation and rigid transforms.

It doesn't support shearing for now (see #90 ) but I'm happy to add that later. 

I would like to ask you to take a look at the API. It's more or less a re-implementation of the imglib2 / clij2 API. But if it makes sense, I'm happy to add additional optional parameters or functions to make it compatible to stuff in snake county. ;-) 

Some interoperability is already available for scikit-image: The `cle.affine_transform()` function takes an `skimage.transform.AffineTransform` as `transform` parameter and converts it internally.

You can take a look at API and its usage in [this notebook](https://github.com/clEsperanto/pyclesperanto_prototype/blob/affine_transforms2/demo/transforms/affine_transforms.ipynb).

Thanks a lot in advance!


tlambert03:
This is super exciting @haesleinhuepf.  Very promising, and I suspect affine transforms will be one of the most widely used parts of this library.

> I would like to ask you to take a look at the API. It's more or less a re-implementation of the imglib2 / clij2 API. But if it makes sense, I'm happy to add additional optional parameters or functions to make it compatible to stuff in snake county. ;-)

As mentioned in https://github.com/clEsperanto/pyclesperanto_prototype/issues/90#issuecomment-757487913, I personally would _love_ to see the API here match the existing python ecosystem APIs of [`scipy.ndimage`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.affine_transform.html) and [`cupyx.scipy.ndimage`](https://docs.cupy.dev/en/stable/reference/generated/cupyx.scipy.ndimage.affine_transform.html)... (which have made attempts to match each other).  I suspect reimplementing the imglib2/clij2 API here makes it easier to generate scripts that go between ImageJ and Python... but in the interest of building bridges, I feel like we need to "speak snake in snake country", "coffee in coffee country" and let the bridge be the compatibility layer in between (rather than asking one of ecosystems to get on board using the language of the other ecosystem).  There are a lot of users of the scipy and cupy APIs, and a cl-based drop-in replacement would be a really amazing contribution.

I know you've had the discussion about explicitly naming axes in parameters elsewhere, but I definitely agree strongly with @jni in https://github.com/clEsperanto/pyclesperanto_prototype/issues/49#issuecomment-749986234 that parameter names should not have any kind of implicit order.  so I don't love `angle_around_z_in_degrees`, but I do like how [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.rotate.html) and [cupy](https://docs.cupy.dev/en/stable/reference/generated/cupyx.scipy.ndimage.rotate.html) do `rotate(input, angle: float, axes: Tuple[int, int])` (as mentioned in https://github.com/clEsperanto/pyclesperanto_prototype/issues/90#issuecomment-757467665)

so, for the sake of completeness:


I know all this takes more work, and I'm happy to start getting more involved here (after tying up some other things), but that would be my vote.

jni:
What @tlambert03 said. 😄 

Note also that in scikit-image/scikit-image#3544, I implemented that `np.array(Transform)` gives back the matrix, and I'm working right now on pulling that change out of that PR so that it can get merged independently. So that might "lighten the load" of what you have to do to make things compatible. 😊

haesleinhuepf:
Alright, in order to do this, I need to wait for another PR to be merged first ( #92 ). Maybe I merge them all together.

Nevertheless, actually basic compatibility can be achieved quite easily, even though API fine tuning might be needed in the future. See #93

PR #93, opened 2021-01-11
Improve interoperability with scipy

haesleinhuepf:
This is for @tlambert03 @jni ;-)
 
I extended the API to be compatible to both worlds (clij and scipy). I added parameters such as `axes` and `angle` to `rotate` and `matrix` to `affine_transform`. Furthermore, I added the aliases `zoom` forwarding to `scale` and `shift` forwarding to `translate`. Also those two have the right parameters to be compatible to scipy (to some degree). I'm considering to make them independent functions, and no aliases, that would mean doubled functionality, but it might be easier to maintain. What do you think?

More sophisticated parameters such as `mode{‘reflect’, ‘constant’, ‘nearest’, ‘mirror’, ‘wrap’}, optional` in zoom and `order` in `affine_transform` are a bit above my abilities. PRs welcome ;-)

Check out [the notebook for details](https://github.com/clEsperanto/pyclesperanto_prototype/blob/improve_interoperability_with_scipy/demo/transforms/affine_transforms.ipynb) (edit: link fixed)

Teaser:
image


haesleinhuepf:
Put this in "Draft" state again, because I'd like to change the API as discussed in https://github.com/clEsperanto/pyclesperanto_prototype/issues/49#issuecomment-757966344

Issue #94, opened 2021-01-11
Use inspect.signature instead of getfullargspec


haesleinhuepf:
Use `inspect.signature` instead of `getfullargspec` as it's the more recent API.

Affected files:
https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/pyclesperanto_prototype/_tier0/_plugin_function.py

Also in related project:
https://github.com/clEsperanto/napari_pyclesperanto_assistant/blob/master/napari_pyclesperanto_assistant/_scriptgenerators/_JythonGenerator.py#L48

jni:
Closed by #95

PR #95, opened 2021-01-12
Fix/Simplify `plugin_function`

tlambert03:
This PR simplifies and fixes some bugs in the heavily used `plugin_function` decorator.
It moves to the more modern `inspect.Signature` API, and takes advantage of the [`Signature.bind` method](https://docs.python.org/3/library/inspect.html#inspect.Signature.bind) to accomplish in a single line what most of the previous implementation was trying to do.

Importantly, this is fixing a big bug in the current implementation, which is that position or keyword arguments were not working.  as an explanation:
```python
def f(x):
    print(x)

# in python a positional argument can also be called as a keyword argument:
f('hi')  # prints "hi"
f(x='hi')  # also prints "hi"
# this is super important because it lets you call a function in many different
# and dynamic ways:
keyword = 'x'
f(**{keyword: 'hi'})  # also prints "hi"
```

the current `plugin_function` breaks this functionality, meaning that functions must be called in a very specific way, or they will break:
```python
from pyclesperanto_prototype import absolute
import numpy as np
r = np.random.rand(10,10) - 0.5

absolute(r)  # works fine
absolute(source=r)  # raises TypeError
```
```pytb
pyclesperanto_prototype/pyclesperanto_prototype/_tier0/_execute.py in execute(anchor, opencl_kernel_filename, kernel_name, global_size, parameters, prog, constants, image_size_independent_kernel_compilation)
    211             arguments.append(np.array([value], np.float32))
    212         else:
--> 213             raise TypeError(
    214                 f"other types than float and int aren`t supported yet for parameters {key} : {value} . \n"
    215                 f"function {kernel_name}"

TypeError: other types than float and int aren`t supported yet for parameters src : [[ 0.42569358  0.41798918  0.487596    0.21005139 -0.19382348 -0.40420797
```

this PR fixes that...

however, I'm still having a bit of an issue with tests, because there was one more thing I'm struggling to figure out here.  The `output_creator` function may be one of many of the functions in `_create.py`, however each of those has a very different signature (some take `*args`, some do not) ... so there too, extra logic is required in `plugin_function` to call the `output_creator` in _just_ the right way depending on the context.  Can you help me understand the intent there?  Should `output_creator` usually just receive the positional arguments provided to the function?  something different?


haesleinhuepf:
Awesome @tlambert03 , the new code is so short! :heart:

Closes #94 

I'll send a PR to your repo for fixing the failing tests...

tlambert03:
thanks!  This is mergeable now... I still think there's some room for improvement on the `output_creator` functions.  Specifically I don't like that we have to call it like this `output_creator(*bound.args[:len(args)])` (because `len(args)` there is dependent on how the outer function was _called_... and that's a little wonky.  I think the fix would require modifying the create functions themselves to have a unified calling signature, but I'm not sure how widely those functions are used elsewhere, so not touching that here.

haesleinhuepf:
> I don't like that we have to call it like this `output_creator(*bound.args[:len(args)])` (because `len(args)` there is dependent on how the outer function was _called_...

Yes! I also don't like that. I struggled hard to make it work earlier and would love to come up with a better solution. You should see the solution on Java side.... It's sooo much worse ;-)

I will continue thinking about it.

Thanks for this PR!

PR #96, opened 2021-01-12
Advanced segmentation2

haesleinhuepf:
This adds a convenience function `voronoi_otsu_labeling` for a workflow I often use: Thresholding + Spot detection + voronoi labeling - similar to a seeded watershed

### DIFF
--- a/pyclesperanto_prototype/_tier9/_voronoi_otsu_labeling.py
+++ b/pyclesperanto_prototype/_tier9/_voronoi_otsu_labeling.py
@@ -1,0 +1,64 @@
+from .. import minimum_of_all_pixels, maximum_of_all_pixels
+from .._tier0 import pull
+from .._tier1 import greater_constant
+
+import numpy as np
+
+from .._tier0 import plugin_function
+from .._tier0 import Image
+from .._tier3 import histogram
+
+@plugin_function(categories=['label', 'in assistant'], priority=1)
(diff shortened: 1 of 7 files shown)


pyclesperanto_prototype/_tier9/_voronoi_otsu_labeling.py
```diff
@@ -0,0 +1,64 @@
...
+from .._tier1 import greater_constant
+
+import numpy as np
+
+from .._tier0 import plugin_function
+from .._tier0 import Image
+from .._tier3 import histogram
+
+@plugin_function(categories=['label', 'in assistant'], priority=1)
+def voronoi_otsu_labeling(source : Image, label_image_destination : Image = None, spot_sigma : float = 2, outline_sigma : float = 2):
```
jni, line 12: To me, this sort of function, which composes other clesperanto functions but does not use its own kernel at all, belongs more in a third-party package (something like [scikit-image-contrib](https://github.com/scikit-image/scikit-image-contrib)) than in pyclesperanto. Does that make sense to you @haesleinhuepf? But, this reorg can come later.

jni (approved):

pyclesperanto_prototype/_tier9/_voronoi_otsu_labeling.py
haesleinhuepf: I see your point. However, this sort of functions are the majority of functions in clesperanto ;-)

Issue #54, opened 2020-12-06
OCLArray and OCLImage


haesleinhuepf:
Hey Talley @tlambert03 ,

I was just [diving](https://github.com/clEsperanto/pyclesperanto_prototype/commit/075250c8391852acd53b5f8038efcd8eeea9473f) a bit into [your](https://github.com/clEsperanto/pyclesperanto_prototype/commit/7073f5f979686de76d6237ace0cab0f8b8e944bf#diff-4eaddb7a6b27bebf0d613f87c254325358688455359f5417d3aebd119972e753) cl.Image code. For affine transforms & friends it's crucial to have cl.Images supported. 
I think we considered refactoring that part earlier and I would have some time to do it in the next weeks. I just wanted to ask you about your opinion/suggestions. I think it makes sense to implement an OCLImage as counter part to [OCLArray](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/pyclesperanto_prototype/_tier0/_pycl.py#L237). Do you agree? Furthermore, I was wondering if there might be things in OCLArray and future OCLImage that could be improved from your perspective. For example having numpy compatibility https://github.com/clEsperanto/pyclesperanto_prototype/issues/33 in mind.

Thanks for your thoughts and hints!

Cheers,
Robert

tlambert03:
not sure what exactly you mean by "as a counter part" ... but if you mean to follow that same `_wrap_OCLArray` pattern, then I'm not sure.  I vaguely recall being not super fond of that pattern with OCLArray, but it's been a while and I need to remind myself.  It could be just the thing :joy: 
I do agree we want `Image` and `Array` to have very similar APIs, so my preliminary Image (and or the OCLArray) code will definitely need some modifications to unify them

I also agree we should maintain as much compatibility with the numpy API as possible (worth reviewing [numpy's dispatch mechanism](https://numpy.org/doc/stable/user/basics.dispatch.html#basics-dispatch)) ...

haesleinhuepf:
> but if you mean to follow that same `_wrap_OCLArray` pattern, then I'm not sure.

An alternative would be implementing something like 'OCLArray.get_as_image : cl.Image'. I think that would allow efficient implementation of all the aspects I would need the cl.Images for....

tlambert03:
I'd say at this point, go with whatever you need to get the job done.  then we can refactor later as the "needs" become more obvious

jni:
[...]

PR #100, opened 2021-03-21
bugfix: cl.Image returned misleading shape

### LINKED ISSUE #99: kernels using cl.Image as parameter don't compile on NVidia / Windows
Dear all, dear future-self,

interpolation for affine transforms is currently not available because of an error while OpenCL compilation. I dived a bit into it and I'm stuck. I'm documenting everything here in case you or me pick it up later.

In very short: Setting `linear_interpolation=True` in `cle.resample` makes pyclesperanto pass an cl.Image to the kernel instad of a cl.Buffer. This leads to an unspecific OpenCL compilation error: `RuntimeError: clBuildProgram failed: <unknown error -9999>`

## To reproduce

This example script for resampling t1-head.tif (ImageJ example image) can be used to reproduce the error:
```
import pyclesperanto_prototype as cle

cle.select_device('RTX')

import numpy as np
import matplotlib
import matplotlib.pyplot as plt
from skimage.io import imshow, imread

# Laod example data
np_array = imread('C:/structure/data/t1-head.tif')

# push it to GPU memory
input_image = cle.push_zyx(np_array)

cle.imshow(input_image)

res = cle.resample(input_image, factor_x=2, linear_interpolation=True)

cle.imshow(res)
```

It fails in the second last line with an error similar to:
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-1-9c5148b1a008> in <module>
     16 cle.imshow(input_image)
     17 
---> 18 res = cle.resample(input_image, factor_x=2, linear_interpolation=True)
     19 
     20 cle.imshow(res)

c:\structure\code\pyclesperanto_prototype\pyclesperanto_prototype\_tier0\_plugin_function.py in worker_function(*args, **kwargs)
     69 
     70         # call the decorated function
---> 71         return function(*bound.args, **bound.kwargs)
     72 
     73     return worker_function

c:\structure\code\pyclesperanto_prototype\pyclesperanto_prototype\_tier1\_resample.py in resample(source, destination, factor_x, factor_y, factor_z, linear_interpolation)
     74     }
     75 
---> 76     execute(__file__, '../clij-opencl-kernels/kernels/affine_transform_' + str(len(destination.shape)) + 'd' + kernel_suffix + '_x.cl',
     77             'affine_transform_' + str(len(destination.shape)) + 'd' + kernel_suffix, destination.shape, parameters)
     78 

c:\structure\code\pyclesperanto_prototype\pyclesperanto_prototype\_tier0\_execute.py in execute(anchor, opencl_kernel_filename, kernel_name, global_size, parameters, prog, constants, image_size_independent_kernel_compilation)
    235         # print("\n".join(defines))
    236 
--> 237         prog = OCLProgram.from_source("\n".join(defines))
    238 
    239         # Todo: the order of the arguments matters; fix that

c:\structure\code\pyclesperanto_prototype\pyclesperanto_prototype\_tier0\_pycl.py in from_source(cls, source)
    177     @lru_cache(maxsize=128)
    178     def from_source(cls, source):
--> 179         return cls(src_str=source)
    180 
    181 

c:\structure\code\pyclesperanto_prototype\pyclesperanto_prototype\_tier0\_pycl.py in __init__(self, file_name, src_str, build_options, dev)
    161         self._kernel_dict = {}
    162         super().__init__(self._dev.context, src_str)
--> 163         self.build(options=build_options)
    164 
    165 

c:\users\rober\miniconda3\lib\site-packages\pyopencl\__init__.py in build(self, options, devices, cache_dir)
    535 
    536             from pyopencl.cache import create_built_program_from_source_cached
--> 537             self._prg, was_cached = self._build_and_catch_errors(
    538                     lambda: create_built_program_from_source_cached(
    539                         self._context, self._source, options_bytes, devices,

c:\users\rober\miniconda3\lib\site-packages\pyopencl\__init__.py in _build_and_catch_errors(self, build_func, options_bytes, source)
    583         # Python 3.2 outputs the whole list of currently active exceptions
    584         # This serves to remove one (redundant) level from that nesting.
--> 585         raise err
    586 
    587     # }}}

RuntimeError: clBuildProgram failed: <unknown error -9999> - clBuildProgram failed: <unknown error -9999> - clBuildProgram failed: <unknown error -9999>

Build on <pyopencl.Device 'GeForce RTX 2080 Ti' on 'NVIDIA CUDA' at 0x2097ef4fad0>:
```

## What I've tried.
I disabled kernel caching [as suggested on Stackoverflow](https://stackoverflow.com/questions/22366517/pyopencl-runtimeerror-clbuildprogram-failed-invalid-build-options). I saw no effect, but this code snippet might be helpful while debugging anyway:
```
import os
os.environ['PYOPENCL_NO_CACHE'] = '1'
```

I also shortened the applied affine_transforms kernel down to the following and still got the compiler error
```
__kernel void affine_transform_3d_interpolate(
    IMAGE_input_TYPE input,
    IMAGE_output_TYPE output,
    IMAGE_mat_TYPE mat
    ) {

}
```

However, when running this, there is another kernel executed for copying the buffer to an image. The copy kernel with an image parameter compiles and executes nicely. When trying to copy the image back (for debugging), I get the same error as above. I assume it has something to do with the first copy operation (it may fail silently) and/or the way how images are created. On Java side we had some issues with some GPUs about channel types. There, we use [R only.](https://github.com/clij/clij-core/blob/master/src/main/java/net/haesleinhuepf/clij/CLIJ.java#L329). 

The error happened on an NVidia RTX 2080 TI, Windows 10 (which works fine in Java). In the same computer, the same code executes on the Intel HD 620. 

Good luck ;-)

haesleinhuepf:
Hey Talley @tlambert03 ,

I know, this is a tricky PR, but I'd like to hear your opinion, because you worked with `cl.Image` earlier. Instead of handling `cl.Image` objects as image parameters directly, I'm wrapping it now into a `_OCLImage`, because:

Our way of using `cl.Image` objects lead to crashes on Nvidia (#99) and to kernels mixing up x and z. I managed to fix this by introducing an `_OCLImage` class (as discussed in #54 ). Basically, the only thing this class does is [changing the order of entries in `image.shape`](https://github.com/clEsperanto/pyclesperanto_prototype/compare/bugfix_resample_interpolation?expand=1#diff-6e922567ae5b27f09b7f0418c6b23f8b6e0cd510748cdece377a811016f5988fR332). It feels like a workaround but I didn't see any other way for fixing this right now. I'm also not sure why this is not necessary with `OCLArray`.

With this change, sampler-based interpolation eventually works while applying affine transforms (this bug was breaking my heart for some time). Furthermore, as we're using `_OCLImage` internally only, we should be fine if we change it later. I think we should either unify the API of `OCLImage` and `OCLArray` or get rid of both at some point. Furthermore, I'm not fully aware of why we need `OCLArray` anyway. Both disappear long term anyway when we switch to the CLIc backend. Long-term.

To make a long story short, this closes #99 and is a partial answer to #54 .

Thanks!

Cheers,
Robert

### DIFF
--- a/docs/release_notes.md
+++ b/docs/release_notes.md
@@ -1,7 +1,8 @@
 # upcoming release
 
 ## Bug fixes
-* Turned of interpolation in `imshow`
+* Turned off interpolation in `imshow`
+* `linear_interpolation=True` caused crashes on NVidida GPUs (#99).
 
 <a name="#070"></a>
 # 0.7.0 - Jan 14th 2021
--- a/pyclesperanto_prototype/_tier0/_cl_image.py
+++ b/pyclesperanto_prototype/_tier0/_cl_image.py
@@ -37,8 +37,8 @@ def empty_image(ctx, shape, dtype, num_channels=1, channel_order=None):
     res = cl.Image(ctx, mem_flags, fmt, shape=shape[::-1])
     res.dtype = dtype
     res.num_channels = num_channels
-    from ._pycl import OCLImage
-    return OCLImage(res)
+    from ._pycl import _OCLImage
+    return _OCLImage(res)
 
 
 def empty_image_like(arr, ctx=None):
--- a/pyclesperanto_prototype/_tier0/_cl_image.py
+++ b/pyclesperanto_prototype/_tier0/_cl_image.py
@@ -37,7 +37,8 @@ def empty_image(ctx, shape, dtype, num_channels=1, channel_order=None):
     res = cl.Image(ctx, mem_flags, fmt, shape=shape[::-1])
     res.dtype = dtype
     res.num_channels = num_channels
-    return res
+    from ._pycl import OCLImage
+    return OCLImage(res)
 
 
 def empty_image_like(arr, ctx=None):
--- a/pyclesperanto_prototype/_tier0/_execute.py
+++ b/pyclesperanto_prototype/_tier0/_execute.py
@@ -5,7 +5,7 @@ from pathlib import Path
 import numpy as np
 
 import pyopencl as cl
-from ._pycl import OCLProgram, OCLImage
+from ._pycl import OCLProgram, _OCLImage
 
 if not os.getenv("PYOPENCL_COMPILER_OUTPUT"):
     import warnings
@@ -159,7 +159,7 @@ def execute(anchor, opencl_kernel_filename, kernel_name, global_size, parameters
                 defines.extend(SIZE_HEADER.format(**params).split("\n"))
 
 
-        elif isinstance(value, OCLImage):
+        elif isinstance(value, _OCLImage):
 
             if value.dtype != np.dtype("float32"):
                 raise TypeError(
--- a/pyclesperanto_prototype/_tier0/_execute.py
+++ b/pyclesperanto_prototype/_tier0/_execute.py
@@ -5,7 +5,7 @@ from pathlib import Path
 import numpy as np
 
 import pyopencl as cl
-from ._pycl import OCLProgram
+from ._pycl import OCLProgram, OCLImage
 
 if not os.getenv("PYOPENCL_COMPILER_OUTPUT"):
     import warnings
@@ -159,7 +159,7 @@ def execute(anchor, opencl_kernel_filename, kernel_name, global_size, parameters
                 defines.extend(SIZE_HEADER.format(**params).split("\n"))
 
 
-        elif isinstance(value, cl.Image):
+        elif isinstance(value, OCLImage):
 
             if value.dtype != np.dtype("float32"):
                 raise TypeError(
@@ -182,7 +182,7 @@ def execute(anchor, opencl_kernel_filename, kernel_name, global_size, parameters
             else:
                 size_parameters = ""
 
-            arguments.append(value)
+            arguments.append(value.data)
 
             if "destination" in key or "output" in key or "dst" in key:
                 type_name = "__write_only image" + str(ndim) + "d_t"
--- a/pyclesperanto_prototype/_tier0/_pycl.py
+++ b/pyclesperanto_prototype/_tier0/_pycl.py
@@ -326,7 +326,7 @@ def _wrap_OCLArray(cls):
 
 OCLArray = _wrap_OCLArray(array.Array)
 
-class OCLImage:
+class _OCLImage:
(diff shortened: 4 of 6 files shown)


tlambert03:
I think the solution here might just be to not reverse the shape in the output of `empty_image`:

https://github.com/clEsperanto/pyclesperanto_prototype/blob/b717b0a321551b0ce8ae19cf8bef337add70809f/pyclesperanto_prototype/_tier0/_cl_image.py#L37

(the new _OCLWrapper is just undoing that basically)

That `_cl_image` code was written assuming it would be the only thing in the pipeline swapping numpy-to-GL shapes, but this library has a number of places that do that (search code `::-1`).  So we need to be careful to keep track of it.  In this case, I think the conflict might be that the shape is also getting swapped in the destination in `execute` here:

https://github.com/clEsperanto/pyclesperanto_prototype/blob/b717b0a321551b0ce8ae19cf8bef337add70809f/pyclesperanto_prototype/_tier0/_execute.py#L229

so, we don't need to swap in `empty_image` anymore

haesleinhuepf:
Thanks for taking a look Talley!

> not reverse the shape in the output of `empty_image`:

Yes, that's what I tried first when I started. Unfortunately, then all code that accesses the shape of those images would have to be changed then. Thus, I decided to do the wrapper-trick. 

> but this library has a number of places that do that (search code ::-1). So we need to be careful to keep track of it.

When crafting the new backend, we should figure out if there is a way for transparent xyz-zyx conversion. I'm afraid we need to change the dimension order inside the OpenCL code. Which migth be tricky but be possible...

tlambert03:
> Thus, I decided to do the wrapper-trick.

I see... so the wrapper trick is basically saying, "whenever pyclesperanto asks for the shape, give it this, but whenever pyopencl asks for the shape, give it this other thing"?

that does make me cringe a little... but I'd say go for whatever you need to get yourself unblocked!  I'm ok with that patch if it lets you move ahead with more important things.  Maybe leave a comment directly in the code back to these two issues for anyone who stumbles on them later?

haesleinhuepf:
> that does make me cringe a little...

Me too! It's eventually the prove that the clij-opencl kernels are incompatible with the python way of handling dimensions. We knew this earlier, but here it's quite obvious: xyz is not zyx ✌️

PR #106, opened 2021-04-17
Improve type support

haesleinhuepf:
With this PR we introduce support for int(8/16/32/64) and uint(8/16/32/64) images and buffers. Per default, operations resulting in binary image will produce uint8 output now and operations producing label images will generate uint32 images.

This is a backwards compatibility breaking change.


haesleinhuepf:
This needs to be tested on all platforms

PR #126, opened 2021-07-24
support _repr_png_

haesleinhuepf:
Hi @tlambert03 and/or @jni,

I'd love to hear what you think about this PR. It's inspired by [Pillow](https://github.com/python-pillow/Pillow/commit/05fe86654cb079b5b77d077e053bf5ee6d2b6f5b). It's the first time, I'm dealing with something like "optional" dependencies, i.e. [image_attendant](https://pypi.org/project/image-attendant/). I don't want clesperanto depend on it, but if it's installed, the feature shown below would be cool. Thus, I was wondering what you think.

It will allow us to see processing results immediately in jupyter notebooks:

image

The "old default" behaviour can be achieved by pulling:

image

It furthermore supports 3D data similar to `cle.imshow`:
image

Cheers,
Robert

### DIFF
--- a/pyclesperanto_prototype/_tier0/_pycl.py
+++ b/pyclesperanto_prototype/_tier0/_pycl.py
@@ -568,6 +568,17 @@ def _wrap_OCLArray(cls):
     #  __rshift__(x1, x2)
     # and, or, xor
 
+    try:
+        # support default output as image in jupyter notebooks
+        import image_attendant as ia
+
+        def _repr_png_(self):
+            return ia.compress(self.get(), 'png')
+
+        cls._repr_png_ = _repr_png_
+    except ImportError:
+        pass
+
     for f in ["dot", "vdot"]:
         setattr(cls, f, wrap_module_func(array, f))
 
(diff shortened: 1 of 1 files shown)


tlambert03:
sure, a `repr_png` is a nice thing to add in certain cases. but I definitely wouldn't depend on image_attendant for it.   just make your own `BytesIO`... here's an example of napari's _repr_png_:
https://github.com/napari/napari/blob/d6bc683b019c4a3a3c6e936526e29bbd59cca2f4/napari/utils/notebook_display.py#L54-L73

I'm also luke warm on using a png only for the repr of an OCLArray... where one very often wants to get shape/dtype, etc... (and I don't think having to call `.get()` is a good solution).  I know that's cool that one can modify the `repr` of an object in a notebook, but using the explicit `imshow()` is also a fine solution in my opinion.  If you really want the image as well, I'd look into the more complex `_ipython_display_` or even `_repr_html_` [methods](https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display), where you could accompany the image with text (like shape, dtype, nbytes, etc...).  like how [dask does](https://github.com/dask/dask/blob/212c1058c19fa543d60d9a294f8f9feda3e1c9dc/dask/array/core.py#L1406)

haesleinhuepf:
That's great pointers @tlambert03, thanks!

PR #131, opened 2021-08-11
Fix pyopencl compatibility

haesleinhuepf:
Hi Talley @tlambert03 ,

this should fix #130 ; and might be important towards #115 . As we can call pyopencl's complex-type-compatible operations.

I'm not sure why I had to implement `__mul__` and friends, but tests containing `cl_array * np_array` didn't pass without...

closes #130 

Let me know what you think!

Best,
Robert

PR #132, opened 2021-08-15
Improve indexing

haesleinhuepf:
Hi Talley @tlambert03 ,

following up our discussion on https://github.com/clEsperanto/pyclesperanto_prototype/pull/131#issuecomment-898972490, I managed to make OCLArrays writable from napari 🥳  With that, raising a custom error becomes obsolete.

I also implemented that __getitem__ asked for a single element return int or float now.

Let me know what you think!

Best,
Robert

### DIFF
--- a/pyclesperanto_prototype/_tier0/_pycl.py
+++ b/pyclesperanto_prototype/_tier0/_pycl.py
@@ -341,8 +341,10 @@ class OCLArray(array.Array, np.lib.mixins.NDArrayOperatorsMixin):
             return power_images(temp, x2, x1)
 
     def __setitem__(self, index, value):
-        if isinstance(index, tuple) and index and isinstance(index[0], np.ndarray):
-            if len(index) == len(self.shape) and len(index[0]) > 1:
+        if isinstance(index, list):
+            index = tuple(index)
+        if isinstance(index, (tuple, np.ndarray)) and index[0] is not None and isinstance(index[0], (tuple, list, np.ndarray)):
+            if len(index) == len(self.shape) and len(index[0]) > 0:
                 # switch xy in 2D / xz in 3D, because clesperanto expects an X-Y-Z array;
                 # see also https://github.com/clEsperanto/pyclesperanto_prototype/issues/49
                 index = list(index)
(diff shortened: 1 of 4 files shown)


haesleinhuepf:
I also just noted, that github desktop added you as coauthor for commits all the time. Sorry for that 🐰 I turned that off now

pyclesperanto_prototype/_tier0/_pycl.py
```diff
@@ -340,13 +340,49 @@ def __ipow__(x1, x2):
             from .._tier1 import power_images
             return power_images(temp, x2, x1)
 
-    def __getitem__(self, index):
-        try:
-            return super().__getitem__(index)
+    def __setitem__(self, index, value):
+        if isinstance(index, tuple) and isinstance(index[0], np.ndarray):
```
tlambert03, line 344:
```suggestion
        if isinstance(index, tuple) and index and isinstance(index[0], np.ndarray):
```

(potential index error otherwise)

tlambert03 (commented):
hi @haesleinhuepf ... if this gets what you need done I suppose it's fine.

It's starting to get extremely specific and makes some very strong assumptions about the underling data structure, which makes this less useful as a generic array.  but this issue has come up many times elsewhere,  and I know this lib is more about cross-compatibility with clij, so I won't repeat it here :)

I do think there are probably some potential user errors here that will no longer provide useful error messages  (e.g. have you tested what happens if any item in the numpy array is out of range?)

in numpy:

```py
In [7]: x = np.arange(10)

In [8]: x
Out[8]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

In [9]: x[np.array([0,2,2])]
Out[9]: array([0, 2, 2])

In [10]: x[np.array([0,2,20])]
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-13-b31901aa2ef5> in <module>
----> 1 x[np.array([0,2,20])]

IndexError: index 20 is out of bounds for axis 0 with size 10
```

anyway, go for it if it takes you in a direction you like :)

haesleinhuepf:
> I do think there are probably some potential user errors here that will no longer provide useful error messages (e.g. have you tested what happens if any item in the numpy array is out of range?)

Hehe, I haven't checked, but I guess clij and clesperanto do the same here: They fallback to the [closest pixel within the image](https://clij.github.io/clij-docs/faq.html#image_edge_handling) a.k,a, [CLK_ADDRESS_CLAMP_TO_EDGE](https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/sampler_t.html). Thus, I think it's ok right now, even though it might not be the same in numpy.

> go for it if it takes you in a direction you like :)

Yes, I think this advanced us quite a bit in the right direction. All that functionality was there before, but hidden behind clij-like commands. With this PR it becomes more pythonic. Step by step.

Thanks for your feedback!

PR #140, opened 2021-11-05
Fix type issues with basic operations

haesleinhuepf:
It was not possible to 
```
y = x * x.max()
```
with x as an OCL-image due to type-mismatches, e.g. if x was of type uint16


haesleinhuepf:
Merging by hand....

PR #143, opened 2021-12-03
add support for exporting to igraph and to networkx + tests and demos

haesleinhuepf:
Hi Szabolcs @szhorvat,

I eventually found some time to write connectors for exporting adjacency graphs to networkx and igraph. I made two methods

* `to_networkx(adjacency_matrix, centroid_positions)`
* `to_igraph(adjacency_matrix, centroid_positions)`

which return respective graph objects and I'm demonstrating them in [this notebook](https://github.com/clEsperanto/pyclesperanto_prototype/blob/support_igraph_and_networkx/demo/neighbors/ipgraph_networkx.ipynb). 

I was hoping you could take a quick look if the API for these two methods makes sense? If there are other libraries with such export functions, are their APIs similar? Is there some kind of community standard?

Any hint is welcome!

Best,
Robert


ntamas:
HI @haesleinhuepf, @szhorvat asked me to chime in and take a look at the Python implementation (I am one of the core developers of `igraph` and `python-igraph`). I'll add a few comments to the commits; feel free to do whatever you want to do with them :)

.github/workflows/test_and_deploy.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           python --version
           conda install -y pyopencl
           python -m pip install --upgrade pip
-          pip install setuptools wheel pytest pytest-cov pytest-benchmark dask
+          pip install setuptools wheel pytest pytest-cov pytest-benchmark dask networkx python-igraph
```
ntamas, line 40: `python-igraph` is now simply `igraph` on PyPI; `python-igraph` will keep on working for a while but it's only a dummy package that depends on `igraph` now. You should install `igraph` directly from PyPI

pyclesperanto_prototype/_tier10/_to_igraph.py
```diff
@@ -0,0 +1,45 @@
...
+    See Also
+    --------
+    ..[1] https://igraph.org/
+    """
+    try:
+        import igraph
+    except ImportError:
+        raise ImportError("igraph is not installed. Please refer to the documentation https://igraph.org/python/")
+
+    igraph_graph = igraph.Graph.Adjacency(np.asarray(adjacency_matrix)[1:,1:].tolist())
```
ntamas, line 33: This is okay but note that `Graph.Adjacency()` now handles NumPy arrays directly (it calls `.tolist()` on the matrix automatically). However, if the matrices are large, this operation can potentially be costly as you need to construct one list for each row of the matrix, so if this is the case, let me know and I'll check whether we could support NumPy arrays in a more efficient manner.

pyclesperanto_prototype/_tier10/_to_igraph.py
```diff
@@ -0,0 +1,45 @@
...
+
+    if centroids is not None:
+        from .._tier1 import transpose_xy
+        centroid_positions = transpose_xy(centroids)
+
+        for n in range(len(igraph_graph.vs)):
```
ntamas, line 39: You can iterate over the vertices directly:

```
for vertex in igraph_graph.vs:
    vertex["x"] = centroid_positions[n][0]
    ...etc...
```

However, it would be more efficient to construct lists of X, Y and Z coordinates for each vertex and then assign them to the graph in a single call (this saves you from constructing `Vertex` objects on the fly in the `for` loop just for the sake of assigning an attribute to them). Something like:

```
if centroids.shape[0] > 2:
    xs, ys, zs = [list(row) for row in centroid_positions.T]
else:
    xs, ys = [list(row) for row in centroid_positions.T]
    zs = [None] * igraph_graph.vcount()
igraph_graph.vs["x"] = xs
igraph_graph.vs["y"] = xs
igraph_graph.vs["z"] = zs
```

pyclesperanto_prototype/_tier10/_to_igraph.py
szhorvat: We really should support numpy arrays for all sorts of dense matrix operations. There should be a direct way to convert them to a C igraph matrix, without going through a Python list of lists.

szhorvat:
Thanks for leaving the comments @ntamas !

pyclesperanto_prototype/_tier10/_to_igraph.py
haesleinhuepf: Hi @ntamas and @szhorvat ,

I'm just playing with [np.nonzero](https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html) and the GPU-counterpart in this project [spots_to_pointlist](https://clij.github.io/clij2-docs/reference_spotsToPointList), They produce lists of coordinates where the adjacency-matrix contains `1`. That should be more efficient to construct the graph from those lists, right?

ntamas: Yes, that should be the preferred way to do it. If you have a 2D NumPy array with 2 columns and N rows, that should do the trick.

haesleinhuepf:
Thanks a lot for your feedback @ntamas and @szhorvat ! 

I just optimized the conversion to igraph by passing an edge list instead of the adjacency matrix using [these three lines](https://github.com/clEsperanto/pyclesperanto_prototype/blob/support_igraph_and_networkx/pyclesperanto_prototype/_tier10/_to_igraph.py#L32-L34) (side note: numpy is faster than its counterpart on the GPU). Also the hint with how to pass the centroids was gold @ntamas, because internally, the `centroids` had the right format already. Graph export should now be much more efficient. 

I furthermore added some convenience functions so that end-users can go from a label image with segmented cells directly to graphs representing n-nearest, proximal or touching neighbor cells. Examples are again in [the notebook](https://github.com/clEsperanto/pyclesperanto_prototype/blob/support_igraph_and_networkx/demo/neighbors/ipgraph_networkx.ipynb).

If you have no further suggestions, I will merge by the end of the week.

Thank you very much again! 🙂

ntamas:
Looks good to me, my only remaining problem is that you should fix the `pip install` line in the Github Actions workflow to refer to `igraph` instead of `python-igraph` because the latter will stop working in a year or so (well, it will keep on working but you won't get any updates).

haesleinhuepf:
Thanks for reminding me of this @ntamas, I missed that one. 

Big thanks again 🙌

PR #150, opened 2022-01-11
Added support for shearing

pr4deepr:
I've added support for shearing. 

I've verified this in CLIJ. I've used the same transform in the affine transform section of the demo notebook as well. This is for shearing in YZ. 
image

However,  the way I've done it is to specify the shearing angle. Is this a good approach, as everywhere else you specify a shearing factor?
The rotation angle option will be great for deskewing as it removes the burden of calculating the shear factor. Should we specify two different options?

Something like this??

`def shear_y(shear_factor_x, shear_factor_z)
`
`def shear_y_angle(rotation_angle_x, rotation_angle_z)`

Open to suggestions!

Ohh, I've also altered this function: 
` def _3x3_to_4x4(self, matrix)
`


pr4deepr:
Or, we could add a separate method for calculating shear factor 
`def calculate_shear_factor`
and then just use shear factor within the shear functions
`def shear_y(shear_factor_x, shear_factor_z)`

pyclesperanto_prototype/_tier8/_AffineTransform3D.py
```diff
@@ -141,20 +141,88 @@ def center(self, shape, undo : bool = False):
 
         return self
 
-    def shear(self):
-        raise NotImplementedError("Shearing has not been implemented yet. \n"
-                                  "See https://github.com/clEsperanto/pyclesperanto_prototype/issues/90")
+    def shear_z(self,rotation_angle_x: float = 0, rotation_angle_y: float = 0 ):
```
haesleinhuepf, line 144:
```suggestion
    def shear_z(self,angle_x_in_degrees: float = 0, angle_y_in_degrees: float = 0 ):
```

pyclesperanto_prototype/_tier8/_AffineTransform3D.py
```diff
@@ -141,20 +141,88 @@ def center(self, shape, undo : bool = False):
...
+        """Shear image in Z along X and/or Y direction
+           Uses angle to calculate the shear factor
+           Typically used for deskewing in lattice lightsheet microscopy
+
+        Args:
+            rotation_angle_x (float, optional): rotation angle in X. Defaults to 0.
+            rotation_angle_y (float, optional): rotation angle in Y. Defaults to 0.
```
haesleinhuepf, line 151:
```suggestion
            angle_x_in_degrees (float, optional): rotation angle in X in degrees. Defaults to 0.
            angle_y_in_degrees (float, optional): rotation angle in Y in degrees. Defaults to 0.
```

pyclesperanto_prototype/_tier8/_AffineTransform3D.py
```diff
@@ -141,20 +141,88 @@ def center(self, shape, undo : bool = False):
...
+            self
+        """          
+        import math
+        try:
+            shear_factor_xz = 1.0 / math.tan(rotation_angle_x * math.pi / 180)
+        except ZeroDivisionError:
+            shear_factor_xz = 0
+        
+        try:
+            shear_factor_yz = 1.0 / math.tan(rotation_angle_y * math.pi / 180)
```
haesleinhuepf, line 163:
```suggestion
            shear_factor_xz = 1.0 / math.tan(angle_x_in_degrees * math.pi / 180)
        except ZeroDivisionError:
            shear_factor_xz = 0
        
        try:
            shear_factor_yz = 1.0 / math.tan(angle_y_in_degrees * math.pi / 180)
```

pyclesperanto_prototype/_tier8/_AffineTransform3D.py
```diff
@@ -141,20 +141,88 @@ def center(self, shape, undo : bool = False):
...
+    def shear_x(self,rotation_angle_y: float = 0, rotation_angle_z: float = 0 ):
+        
+        import math
+        try:
+            shear_factor_yx = 1.0 / math.tan(rotation_angle_y * math.pi / 180)
+        except ZeroDivisionError:
+            shear_factor_yx = 0
+        
+        try:
+            shear_factor_zx = 1.0 / math.tan(rotation_angle_z * math.pi / 180)
```
haesleinhuepf, line 185:
```suggestion
    def shear_x(self,angle_y_in_degrees: float = 0, angle_z_in_degrees: float = 0 ):
        
        import math
        try:
            shear_factor_yx = 1.0 / math.tan(angle_y_in_degrees * math.pi / 180)
        except ZeroDivisionError:
            shear_factor_yx = 0
        
        try:
            shear_factor_zx = 1.0 / math.tan(angle_z_in_degrees * math.pi / 180)
```

pyclesperanto_prototype/_tier8/_AffineTransform3D.py
```diff
@@ -141,20 +141,88 @@ def center(self, shape, undo : bool = False):
...
+    def shear_y(self,rotation_angle_x: float = 0, rotation_angle_z: float = 0 ):
+        
+        import math
+        try:
+            shear_factor_xy = 1.0 / math.tan(rotation_angle_x * math.pi / 180)
+        except ZeroDivisionError:
+            shear_factor_xy = 0
+        
+        try:
+            shear_factor_zy = 1.0 / math.tan(rotation_angle_z * math.pi / 180)
```
haesleinhuepf, line 207:
```suggestion
    def shear_y(self,angle_x_in_degrees: float = 0, angle_z_in_degrees: float = 0 ):
        
        import math
        try:
            shear_factor_xy = 1.0 / math.tan(angle_x_in_degrees * math.pi / 180)
        except ZeroDivisionError:
            shear_factor_xy = 0
        
        try:
            shear_factor_zy = 1.0 / math.tan(angle_z_in_degrees * math.pi / 180)
```

haesleinhuepf (changes requested):
Hi Pradeep @pr4deepr ,

thanks for adding this. The API looks great!

> The rotation angle option will be great for deskewing as it removes the burden of calculating the shear factor. Should we specify two different options?
> 
> Something like this??
> 
> `def shear_y(shear_factor_x, shear_factor_z) ` `def shear_y_angle(rotation_angle_x, rotation_angle_z)`

I think we should not offer both options. The angle should be the more intuitive way for shearing. However, if you think that makes sense, we could add a comment to the documentation that explains how to get the angle from the shaering factor. I'm just wondering: Can you imagine a scenario, where the user knows the shearing factor, but not the angle?

> Ohh, I've also altered this function:
> def _3x3_to_4x4(self, matrix)

Big thhanks for this! You also removed my embarressing comment. ;-)

I'm suggesting to rename the "angle" parameters, because it's not a rotation angle, it's a shearing angle. ;-) I'm also adding the "in_degrees" because we also have that in the [rotate](https://github.com/clEsperanto/pyclesperanto_prototype/blob/bd378bb76d7400c992dc159eb04a458a205cbd88/pyclesperanto_prototype/_tier8/_AffineTransform3D.py#L57) function and in that way the API is consistent. Do you think that's ok?

Furthermore, a tricky aspect: We should add a test for the new API. Could you think of a simple example for testing the shearing? Maybe something like a 5x5x5 large image stack with all pixels 0 but one pixel set to 1 which we shear in all directions by 45 degrees?

Big thanks again!
Robert

pr4deepr:
> I think we should not offer both options. The angle should be the more intuitive way for shearing. However, if you think that makes sense, we could add a comment to the documentation that explains how to get the angle from the shaering factor. I'm just wondering: Can you imagine a scenario, where the user knows the shearing factor, but not the angle?

I had a look at scikit image just to see how they implement [shear ](https://scikit-image.org/docs/dev/api/skimage.transform.html#skimage.transform.AffineTransform)and realised that they specify the angle (in radians). For our case, angle in degrees is fine and its a good idea to be explicit that its in degrees.
I was using shear in CLIJ, which is why I was mentioning to use shear in pycleperanto. But, it appears that using the angle is convention as far as I can tell. 

`Big thhanks for this! You also removed my embarressing comment. ;-)
`
I'm sure there is a better way than what I did.. Probably worth leaving that in! :)

> Furthermore, a tricky aspect: We should add a test for the new API. Could you think of a simple example for testing the shearing? Maybe something like a 5x5x5 large image stack with all pixels 0 but one pixel set to 1 which we shear in all directions by 45 degrees?
Done! I like this idea of using a 5x5x5 matrix...

Let me know if it makes sense.

Cheers
Pradeep

haesleinhuepf (approved):
❤️

Issue #151, opened 2022-01-14
Issues with shearing 


haesleinhuepf:
Hi @pr4deepr ,

just to let you know, the current implementation of shearing does not well fit compared to CLIJ. I'm going to dive a bit deeper and keep you posted.

In CLIJ we consume a factor (and not an angle). shearXY=1 results in this:
![image](https://user-images.githubusercontent.com/12660498/149516549-e4d5d985-1d7e-4a48-b774-2467f2f2ec54.png)

In pyclesperanto, if I call similar commands the result appears it shears along the wrong axis:

```
shear_angle = 45.0

transform = cle.AffineTransform3D()
shear_transform= transform.shear_x(angle_y_in_degrees=shear_angle)

transformed_image = cle.affine_transform(input_image, transform=shear_transform)

cle.imshow(transformed_image[64])
```
![image](https://user-images.githubusercontent.com/12660498/149516742-835c3b0c-d709-4b1b-9697-6c58c2f1ad4f.png)

Furthermore, shearing by an angle of almost 0 doesn't return the original image, as I would suspect:

```
for shear_angle in [0, 30, 75, 90]:
    transform = cle.AffineTransform3D()
    shear_transform= transform.shear_x(angle_y_in_degrees=shear_angle)

    transformed_image = cle.affine_transform(input_image, transform=shear_transform)

    cle.imshow(transformed_image[64])
```

![image](https://user-images.githubusercontent.com/12660498/149517097-f2d8e1b8-b076-4fe7-a232-3c74de6d877e.png)

Again, I'm investigating... Enjoy your weekend! ;-)

haesleinhuepf:
This was fixed in #152

PR #152, opened 2022-01-14
Shear api convenience

haesleinhuepf:
Hi Pradeep @pr4deepr ,

I'm suggesting to change the shear API to be more intuitive and comparable to CLIJ. You know, datasets in python are ZYX and in java XYZ. I think that's a part of the reason why the former API didn't replicate CLIJ well. See also #151 

Anyway, I'm now suggesting that a function `shear_in_z_plane()` works like the following. If I shear X in the Z-plane with different angles, I see this:
```
for shear_angle in [-30, -15, 0, 15, 30, 90]:
    transform = cle.AffineTransform3D()
    shear_transform= transform.shear_in_z_plane(angle_x_in_degrees=shear_angle)

    transformed_image = cle.affine_transform(input_image, transform=shear_transform)

    cle.imshow(transformed_image[64])
```
image

Hence, a shearing of 0 keeps the original image untouched. Furthermore, shearing with angle 90 degress should cause a warning, or an error even.
image

If I shear Y in the Z plane, it looks like this:
```
for shear_angle in [-30, -15, 0, 15, 30, 90]:
    transform = cle.AffineTransform3D()
    shear_transform= transform.shear_in_z_plane(angle_y_in_degrees=shear_angle)

    transformed_image = cle.affine_transform(input_image, transform=shear_transform)

    cle.imshow(transformed_image[64])
```
image

There's a complete notebook, where you can play with this API:
https://github.com/clEsperanto/pyclesperanto_prototype/blob/shear_API_convenience/demo/transforms/affine_transforms-Copy1.ipynb

Let me know what you think! If this sounds reasonable, I can also make functions for shearing in the X and Y planes.


pr4deepr:
Hi @haesleinhuepf 
    Apologies, I was thinking of CLIJ and didn't realise pyclesperanto will be using Python convention.. I
This axes order of XYZ and YZX confused me a lot when I started figuring out transforms. I refer to when I was implementing shear transforms in Python

image

I do like the API and it is quite simple. I'll comment on the code, as I had a few questions..

Cheers
Pradeep

pyclesperanto_prototype/_tier8/_AffineTransform3D.py
```diff
@@ -157,20 +157,22 @@ def shear_z(self,angle_x_in_degrees: float = 0, angle_y_in_degrees: float = 0 ):
...
-            shear_factor_xz = 1.0 / math.tan(angle_x_in_degrees * math.pi / 180)
+            shear_factor_xz = 1.0 / math.tan((angle_x_in_degrees-90) * math.pi / 180)
         except ZeroDivisionError:
-            shear_factor_xz = 0
+            print("Error1")
+            shear_factor_xz = math.pi
         
         try:
-            shear_factor_yz = 1.0 / math.tan(angle_y_in_degrees * math.pi / 180)
+            shear_factor_yz = 1.0 / math.tan((angle_y_in_degrees-90) * math.pi / 180)
```
pr4deepr, line 166: The way I've been calculating shear factor is by the equation:
`np.tan((90-angle) * np.pi / 180)`
Shouldn't you be using: 
`1.0 / math.tan((angle_x_in_degrees) * math.pi / 180)` ?? 
Whats the logic behind using `angle_y_in_degrees-90` ?

pyclesperanto_prototype/_tier8/_AffineTransform3D.py
```diff
@@ -157,20 +157,22 @@ def shear_z(self,angle_x_in_degrees: float = 0, angle_y_in_degrees: float = 0 ):
...
-            shear_factor_yz = 0
+            print("Error2")
+            shear_factor_yz = math.pi
 
         # shearing
-        self._concatenate(np.asarray([
-            [1, 0, shear_factor_xz, 0],
-            [0, 1, shear_factor_yz, 0],
-            [0, 0 , 1, 0],
+        self._pre_concatenate(np.asarray([
```
pr4deepr, line 172: Why do you use a pre_concatenate for the shear matrices?

pyclesperanto_prototype/_tier8/_AffineTransform3D.py
haesleinhuepf: The logic behind -90 degrees is that an angle of 0 should not shear. Or am I thinking wrong here?
You can see that in the small for-loops of the example notebook.

haesleinhuepf: 90 - angle might do the same thing though...

pyclesperanto_prototype/_tier8/_AffineTransform3D.py
haesleinhuepf: Because we describe a forward transform (from source image to target), but we apply a backward-transfrom (from every pixel in target, we have to find out where the source pixel is). This is equivalent to a .inverse() operation.

Goal is to define an API that is intuitive from the outside (saying, I want shear this and rotate that) and then internally do the math right so that the result image is correct, according to what the user wrote.

pr4deepr:
Hi @haesleinhuepf 

   I've been trying to debug this. 
Does your affine transformation in [affine_transform_3d_x.cl](https://github.com/clEsperanto/clij-opencl-kernels/blob/9374863985160dd80854a3576e9cf863ee9952dd/kernels/affine_transform_3d_x.cl)
automatically correct for pythons axes, where it switches it from zyx to xyz mode??


For me the shear transformation matrix when images are in the axes order `xyz `is the one that works, i.e., the first matrix here:
https://github.com/clEsperanto/pyclesperanto_prototype/pull/152#issuecomment-1013834497

I've included the function that works for me:
```
def shear_in_z(angle_x_in_degrees: float = 0, angle_y_in_degrees: float = 0 ):
    """Shear image in Z-plane along X and/or Y direction
       Uses angle in degrees to calculate the shear
       Tip: Used for lattice lightsheet deskewing. For Janelia lattice, use angle_x_in_degrees and for Zeiss lattice, use angle_y_in_degrees

    Args:
        angle_x_in_degrees (float, optional): shear angle in X in degrees. Defaults to 0.
        angle_y_in_degrees (float, optional): shear angle in Y in degrees. Defaults to 0.
                angle in degrees. To convert radians to degrees use this formula:
                angle_in_deg = angle_in_rad / numpy.pi * 180.0

    Returns:
        self
    """          
    import math
    
    try:
        shear_factor_xz = 1/math.tan((angle_x_in_degrees) * math.pi / 180)
    except ZeroDivisionError:
        print("Error1")
        shear_factor_xz = 0
            
    try:
        shear_factor_yz = 1/math.tan((angle_y_in_degrees) * math.pi / 180)
    except ZeroDivisionError:
        print("Error2")
        shear_factor_yz = 0
    mat = cle.AffineTransform3D()._matrix
    mat1 = (np.asarray([
        [1, 0, shear_factor_xz, 0],
        [0, 1, shear_factor_yz, 0],
        [0, 0, 1, 0],
        [0, 0, 0, 1],
    ]))
    #print(mat1)
    return np.matmul(mat, mat1)

```

With calculating the shear factor, these are the two equations that I've been using. Either :

`math.tan((90-angle_x_in_degrees) * math.pi / 180)
`
OR

`1/math.tan((angle_x_in_degrees) * math.pi / 180)`


I'll link a lattice dataset you can test it on:

https://cloudstor.aarnet.edu.au/plus/s/eAy7Wypjkt3IjtN/download

Shear angle should be 30 degrees and needs deskewing in Y direction

My understand is I have to shear it in Z plane along the Y direction. Unless, I misunderstood this completely.

Cheers
Pradeep

pr4deepr:
jupyter notebook so you can test this:
https://cloudstor.aarnet.edu.au/plus/s/570Ivm8MywBWeIX

pr4deepr:
I just wanted to document this option. 

Another way is to use the[ transforms3d shear API](https://matthew-brett.github.io/transforms3d/reference/transforms3d.shears.html#sadn2mat)
For shear, we can use:
`transforms3d.shears.sadn2aff(angle, direction, normal, point=None)
`
_Matrix for shear by angle along direction vector on shear plane_
`angle `= shear angle in radians
`direction `= shear direction, if its sheared in Y, its [0,1,0] or if its sheared in X its [1,0,0]
`normal `= plane along which shear occurs. For Z, its [0,0,1]
`point `= None means the shear plane is through origin

If angle is 30 degrees, and image sheared in Z-plane along Y direction

shear angle in radians = (90 - 30)* math.pi / 180  = 1.0472

So, for image in the axes order (xyz) the function will be of the form:

shear_matrix = transforms3d.shears.sadn2aff(1.0472, [0,1,0], [0,0,1])

This returns the following shear matrix for shear in YZ:

[1, 0, 0, 0]
[0, 1, 1.732, 0]
[0, 0, 1, 0]
[0, 0, 0, 1]

pyclesperanto_prototype/_tier8/_AffineTransform3D.py
pr4deepr: Thanks
I think I'm still confused, you perform the inverse of the affine transform here, right?
[affine_transform.py](https://github.com/clEsperanto/pyclesperanto_prototype/blob/58e9f2239a31add046d539c4483e3e9798b0886b/pyclesperanto_prototype/_tier8/_affine_transform.py#L75)?
Apologies if this is getting too messy!

haesleinhuepf:
> jupyter notebook so you can test this: https://cloudstor.aarnet.edu.au/plus/s/570Ivm8MywBWeIX

In this notebook there is one thing, that I'm not sure if I understand right. shering in Z mean, according to my interpretation, that I see a shearing in the X-Y image. But this block looks like it's shearing in the YZ plane:
image

You know I _think_ it would be good if "shearing in z plane" would be visible to the user if they look in a Z plane and tune shearing along X and Y...

pr4deepr:
Wasn't sure how to update this pull request, I submitted one here:
https://github.com/clEsperanto/pyclesperanto_prototype/pull/156

haesleinhuepf:
Hi Pradeep @pr4deepr ,

I think I have it now consistent and intuitive. We still had some confusion between the planes, but I now explicitly explain everything in the [new shearing demo notebook](https://github.com/clEsperanto/pyclesperanto_prototype/blob/993bb1f7b139f8fafdd6ee397811793c7eae65af/demo/transforms/shearing.ipynb). I hope the explanation of what the Z-plane is helps others to adopt the API. The notebook also contains a section explaining how clesperanto is compatible with CLIJ. It should be 100% backwards compatible. I'm also demonstrating the transition between a shearing angle and the corresponding shearing factor.

I assume for your deskewing problem, you want to shear along Z in the Y plane. and hence, `shear_in_y_plane(angle_z_in_degress=30)` should do the job.

Let me know what you think!

Thanks,
Robert

pr4deepr:
Hi @haesleinhuepf 
   Ahh, I understand where I was confused! 
Thanks for explaining it clearly in the documentation. 
Is it worth mentioning that the affine matrices are in the form: xyz for the developers. as Python is traditionally in the zyx form. I know you can define it using clesperanto and don't need to manually define it, but I feel this can get lost in translation literally.. :)  Perhaps I missed this in the documentation. 

The API is elegant and I really like the inclusion of determine_bounding_boxes and translation in the code.. 
Apart from calling it directly, is there a checkbox or ability to access this?

Cheers
Pradeep

pr4deepr (approved):

haesleinhuepf:
> determine_bounding_boxes and translation in the code..
> Apart from calling it directly, is there a checkbox or ability to access this?

Yes, it's that "auto-size" checkbox that came with https://github.com/clEsperanto/pyclesperanto_prototype/pull/153

pr4deepr:
Of course, I lost track! 
Thanks..

PR #153, opened 2022-01-15
Auto size affine transforms

haesleinhuepf:
Hi Pradeep @pr4deepr , 

I'm adding the `auto_size` parameter to all transforms (except shear yet) as discussed in our last meeting. I'm super happy with it, thanks for the suggestion! Activating the checkbox in the assistant makes it very easy to use. :-)

I'm just sending this PR as it was your code and in case you wanted to have a quick look. Your code mainly went [here](https://github.com/clEsperanto/pyclesperanto_prototype/blob/932516eb5b9d8bcd36acb47f7e1144d325c010b3/pyclesperanto_prototype/_tier8/_affine_transform.py#L116).

Cheers,
Robert


pr4deepr:
Glad its of use and thanks for letting me know!

PR #156, opened 2022-01-19
Shear api convenience

pr4deepr:
Updated the shear API 
If pyopencl code works on arrays representing x,y,z, then I reckon the affine transforms should be in that order as well as the affine transformation happens at the GPU with the image in x,y,z order.
I've just done the API for `shear_in_z_plane `and updated the jupyter notebook you created. 
Can you check if it works?


haesleinhuepf:
Hey Pradeep @pr4deepr ,

half of it was perfect! However, in the first example in the notebook, `shear_in_z_plane` didn't shear in the z-plane. I'll update the code and let you know in the other PR.

Thanks!
Robert

PR #157, opened 2022-01-25
Deskew

haesleinhuepf:
Hi Pradeep @pr4deepr ,

here comes now the promised `deskew_y()` method. For now just in Y-direction until we are sure that everything is fine. I can then add the X-analogon. I don't see any other use-cases, correct?

I took the opportunity to copy the example dataset from the other repository. I converted it to 8-bit (so that it's a bit smaller) and resaved it as TIF for accessibility reasons. Would you mind confirming that it is ok, that we use this data in this public repository? If the authors are on github, I'm happy to tag their user-names in the release notes. Their names and affiliation are mentioned in a text file next to the image, in the release notes and in the example notebook. Alternatively, I can get some data from our SCAPE. No big deal, just let me know.

https://github.com/clEsperanto/pyclesperanto_prototype/blob/deskew/demo/transforms/deskew.ipynb

Could you also check if the microscopy techniques in the notebook are correctly described? You never know if the computer scientists writes bull***t ;-)

Please let me know what you think!

Thanks!

Best,
Robert


pr4deepr:
Hi @haesleinhuepf 
  I'm going through this and get back to you by tomorrow. Will edit a few things!


Few points:
- We have data skewed in the X-direction from the Janelia lattice microscope, so that can be a good demo dataset too for testing and application of different shear angles. Would it be useful?
- Happy to include the dataset from the other repo in here. Will alter the acknowledgments accordingly. 
This is looking exciting!

Cheers
Pradeep

haesleinhuepf:
> * Would it be useful?

Yes, that would be fantastic! Not sure about GitHub file size limits, but keep it small 😁

> get back to you by tomorrow.

Enjoy the public holiday! 😃

pr4deepr:
I've had a look at this and made changes in the pull request:
https://github.com/clEsperanto/pyclesperanto_prototype/pull/158

Let me know what you think..

haesleinhuepf:
Hi Pradeep @pr4deepr ,

thanks again for the deskew-scaling! You see now here how I adapted it. The crucial line is [this one](https://github.com/clEsperanto/pyclesperanto_prototype/pull/157/files#diff-7265737ae885454532135a2ce143341ff9106bb73a03d9f9817c77f959110f17R55). After the dataset has been sheared and rotated, I'm adapting the voxel size through scaling. At the moment, we are rescaling the dataset in a way, that the voxels are isotropic at the end. These may be huge image stacks but not really oversampled as the focal plane lies between the planes. Anyway, if the data becomes too big, we could use the `scaling_factor` I also introduced, just in case... It's per default 1. I'm showing how to use it in the demo notebook as well:

https://github.com/clEsperanto/pyclesperanto_prototype/blob/7ce62ee284246bc6e40d7a49f9132fb4c4836fae/demo/transforms/deskew.ipynb

I also copied your comment about voxel size up in the notebook and point to the fact that you MUST know voxel size and deskewing angle. Otherwise the reconstruction will just be wrong.

Furthermore, if you want to compare your implementation of the `deskew` with the one that is currently implemented, you can check out this notebook (which I will delete before merging)

https://github.com/clEsperanto/pyclesperanto_prototype/blob/7ce62ee284246bc6e40d7a49f9132fb4c4836fae/demo/transforms/deskew_test.ipynb

Let me know what you think!

Robert

haesleinhuepf:
Awesome. Thanks again @pr4deepr - I couldn't have done that alone!

PR #158, opened 2022-01-27
Deskew added scaling

pr4deepr:
Hi Robert @haesleinhuepf 
   Could you merge this into : https://github.com/clEsperanto/pyclesperanto_prototype/pull/157   ?
Is this how I create a pull request for the above issue? Do you then merge this request to above?

Anyway, the changes:

- Added scaling option for deskewing
- Fixed data description and author affiliations/name
- Added option for deskewing with scaling
- Added scale_z_with_pixel_size function in transformations, so that pixel size can be used to calculating scaling factor and return corresponding scaling matrix. I'm not sure where else to place this function. It was either utilities or transformations..
- Updated notebook with option for scaling

Cheers
Pradeep


pyclesperanto_prototype/_tier8/_deskew_y.py
```diff
@@ -35,5 +44,7 @@ def deskew_y(input_image: Image, output_image: Image = None, angle_in_degrees: f
     # correct orientation so that the new Z-plane goes proximal-distal from the objective.
     transform.rotate(angle_in_degrees=90, axis=0)
 
+    transform.scale_in_z_with_pixel_size(angle_in_degrees,pixel_size_xy=dxy,pixel_size_z=dz)
```
haesleinhuepf, line 47: Hi Pradeep @pr4deepr , just a quick question: instead of scaling with a sheared scaling factor, can we alternatively scale before the shearing? Can this achieve the same?

pr4deepr: Hi @haesleinhuepf  
   Is it just using the scale function?
BTW, I have been testing the RBC_test images and the deskewed images appear squashed when viewing them in napari. 
The dataset is from a Zeiss lattice where the  angle is 30 degrees. 
I'll get you the raw data and corresponding deskewed image from the microscope using proprietary software tomorrow so we can compare..

Cheers
Pradeep

haesleinhuepf: When thinking of an intuitive programming / user interface I think, we should ask the user to input the voxel size during acquisition AND the voxel size of the image that should be reconstrunted. That would be six numbers though...

pyclesperanto_prototype/_tier8/_AffineTransform3D.py
```diff
@@ -250,6 +250,21 @@ def shear_in_x_plane(self,angle_y_in_degrees: float = 0, angle_z_in_degrees: flo
         ]))
         return self
 
+    def scale_in_z_with_pixel_size(self, angle_in_degrees: float = 0, pixel_size_xy: float = 0, pixel_size_z: float = 0 ):
```
haesleinhuepf, line 253: We typically don't have variables called `pixel_size_xy` in clesperanto. We use `pixel_size_x` and `pixel_size_y` separately to be flexible.

pyclesperanto_prototype/_tier8/_deskew_y.py
```diff
@@ -1,10 +1,11 @@
+from numpy import angle
 from .._tier0 import plugin_function
 from .._tier0 import Image
 from .._tier0 import create_none
 
 
 @plugin_function(output_creator=create_none)
-def deskew_y(input_image: Image, output_image: Image = None, angle_in_degrees: float = 30) -> Image:
+def deskew_y(input_image: Image, output_image: Image = None, angle_in_degrees: float = 30, dxy:float =0, dz:float =0) -> Image:
```
haesleinhuepf, line 8: Variable names such as `dxz` and `dz` are no good practice. Consider using descriptive english words for variables.

pyclesperanto_prototype/_tier8/_deskew_y.py
```diff
@@ -16,6 +17,12 @@ def deskew_y(input_image: Image, output_image: Image = None, angle_in_degrees: f
         reconstructed image data with Z-planes in proximal-distal oriental from the objective
     angle_in_degrees: float, optional
         default: 30 degrees
+    dxy: float, optional
+         default: 1 pixel per micron
```
haesleinhuepf, line 21: The default written here is wrong. I agree, it should be 1. However, above 0 is specified. Scaling with value 0 is not a good idea ;-)

pyclesperanto_prototype/_tier8/_utilities.py
```diff
@@ -111,4 +111,4 @@ def shear_factor_to_shear_angle(shear_factor):
     -------
     float
     """
-    return  - math.atan(1.0 / shear_factor) * 180 / math.pi + 90
+    return  - math.atan(1.0 / shear_factor) * 180 / math.pi + 90
```
haesleinhuepf, line 114: What happened here? (-:

haesleinhuepf (commented):
Hi Pradeep @pr4deepr ,

I'm going to merge this and fix some issues with it afterwards. I will for example remove the `cale_in_z_with_pixel_size` function as it's not necessary. You will see the result in the other pull-request #157 .

I'm just adding some comments below to demonstrate, what I'm going to change and why.

Thanks for this! I like how we ping-pong pull-requests. And at the end, there will be an awesome deskewing API in clesperanto. :-)

Best,
Robert

PR #159, opened 2022-01-28
Deskewing fix

pr4deepr:
Hey @haesleinhuepf  

  I had a look at the calculations and fixed two things:

- Rotation angle: As we are rotating about axis 0, it should be rotated around 90 - shearing_angle. 
- Shearing factor calculation: I reverted to the calculation I was using before. Another option is to use transforms3D. 
- Scale factor calculation: I've used the method where it calculates scale factor based on shearing angle and initial voxel z size in micron. I've referred to Talley's code where he used a similar calculation:
https://github.com/tlambert03/napari-ndtiffs/blob/092acbd92bfdbf3ecb1eb9c7fc146411ad9e6aae/napari_ndtiffs/affine.py#L57

Try this on the data I've shared and you should get similar dimensions.. 

Cheers
Pradeep


pyclesperanto_prototype/_tier8/_deskew_y.py
```diff
@@ -41,18 +41,20 @@ def deskew_y(input_image: Image,
...
+    # rotate the stack to get proper Z-planes; rotate 90 - angle around X-axis
+    transform.rotate(angle_in_degrees=90-angle_in_degrees, axis=0)
 
-    # make voxels isotropic, equal to voxel size to raw image in X times scaling_factor.
-    scaling_factor_y = voxel_size_y / voxel_size_x * scaling_factor
-    scaling_factor_z = voxel_size_z / voxel_size_x * scaling_factor
-    transform.scale(scale_x=scaling_factor, scale_y=scaling_factor_y, scale_z=scaling_factor_z)
+    # make voxels isotropic, calculate the new scaling factor for Z after shearing
+    #https://github.com/tlambert03/napari-ndtiffs/blob/092acbd92bfdbf3ecb1eb9c7fc146411ad9e6aae/napari_ndtiffs/affine.py#L57
+    new_dz=math.sin(angle_in_degrees * math.pi/180.0)*voxel_size_z
```
pr4deepr, line 55: Could put this in a function, but didn't want to do it until we've decided on the best approach

pyclesperanto_prototype/_tier8/_utilities.py
```diff
@@ -97,7 +97,11 @@ def shear_angle_to_shear_factor(angle_in_degrees):
     -------
     float
     """
-    return 1.0 / math.tan((90 - angle_in_degrees) * math.pi / 180)
+    try:
+        shear_factor = 1/math.tan((angle_in_degrees) * math.pi / 180)
+    except ZeroDivisionError:
+        shear_factor = 0
+    return shear_factor
```
haesleinhuepf, line 104:
```suggestion
    return 1.0 / math.tan((90 - angle_in_degrees) * math.pi / 180)
```

We should undo this change, because it breaks shearing (see demo/transforms/shearing.ipynb). I will make a suggestion of where to make a change that fixes deskew AND keeps shearing working ;-)

pyclesperanto_prototype/_tier8/_deskew_y.py
```diff
@@ -41,18 +41,20 @@ def deskew_y(input_image: Image,
 
     from ._AffineTransform3D import AffineTransform3D
     from ._affine_transform import affine_transform
-
+    import math
+    
     # shear in the X plane towards Y
     transform = AffineTransform3D()
     transform.shear_in_x_plane(angle_y_in_degrees=angle_in_degrees)
```
haesleinhuepf, line 48:
```suggestion
    transform.shear_in_x_plane(angle_y_in_degrees=90 - angle_in_degrees)
```
Does that make sense here?

pyclesperanto_prototype/_tier8/_deskew_y.py
haesleinhuepf: I think it's ok here. We only need it in this one place, right?

haesleinhuepf (changes requested):
Hi Pradeep @pr4deepr ,

I think we're almost there. WE need to undo one of your suggestions, because it breaks shearing. However, I'm making a suggestion of making the same change in a different place  so that both deskewing and shearing should work as expected.

Let me know if it makes sense!

Best,
Robert

pyclesperanto_prototype/_tier8/_deskew_y.py
pr4deepr: Actually, we will need it for the deskew_x function as well. We will change line 56:
scale_factor_z=(new_dz/`voxel_size_y`)*scaling_factor

to 

scale_factor_z=(new_dz/`voxel_size_x`)*scaling_factor

Technically pixel size in x and y are the same..
I'll get a lattice dataset skewed in X direction next week so we can have that example too.

pyclesperanto_prototype/_tier8/_deskew_y.py
pr4deepr: Haha, it make so much sense to add it here. 
Both the shear and the rotate have consistent angles as well, as in they both use` 90 - angle`

pr4deepr (commented):
Your suggestions keep it consistent and doesn't break anything in the pyclesperanto backend!!
This is looking great.. 
Will add deskew_x soon. Shouldn't be too hard...

haesleinhuepf:
Awesome thanks!

> Will add deskew_x soon. Shouldn't be too hard.

I would prefer this solution. That would be consistent with the clesperanto API, where we also have `maximum_x_projection`, `maximum_y_projection` etc.

I'll merge this PR here soon, 157 as well, and look forward to another PR with deskew_x. There is no hurry btw. We could also easily add it with a later release.

Have a great weekend! 🌞

pr4deepr:
Thanks, you have a great weekend too!

PR #161, opened 2022-01-29
memory allocation optimization

elsandal:
Removed redundant variable to lower memory allocation. float_arr is not needed, we can return without creating a new variable. We can reduce the memory requirements to 1x the image size.

### DIFF
--- a/pyclesperanto_prototype/_tier0/_push.py
+++ b/pyclesperanto_prototype/_tier0/_push.py
@@ -34,8 +34,7 @@ def push(any_array):
     if hasattr(any_array, 'shape') and hasattr(any_array, 'dtype') and hasattr(any_array, 'get'):
         any_array = np.asarray(any_array.get())
 
-    float_arr = any_array.astype(np.float32)
-    return Backend.get_instance().get().from_array(float_arr)
+   return Backend.get_instance().get().from_array(np.float32(any_array))
 
 
 def push_zyx(any_array):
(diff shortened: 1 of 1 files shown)


pyclesperanto_prototype/_tier0/_push.py
```diff
@@ -34,8 +34,7 @@ def push(any_array):
     if hasattr(any_array, 'shape') and hasattr(any_array, 'dtype') and hasattr(any_array, 'get'):
         any_array = np.asarray(any_array.get())
 
-    float_arr = any_array.astype(np.float32)
-    return Backend.get_instance().get().from_array(float_arr)
+   return Backend.get_instance().get().from_array(np.float32(any_array))
```
haesleinhuepf, line 37:
```suggestion
    return Backend.get_instance().get().from_array(np.float32(any_array))
```
Hi @elsandal ,

I'm just adding a space here, because the tests were failing. The indentation was wrong. ;-)

pyclesperanto_prototype/_tier0/_push.py
```diff
@@ -44,4 +43,4 @@ def push_zyx(any_array):
         "Deprecated: `push_zyx()` is now deprecated as it does the same as `push()`.",
         DeprecationWarning
     )
-    return push(any_array)
\ No newline at end of file
+    return push(any_array)
```
haesleinhuepf, line 46:
```suggestion
    return push(any_array)

```

haesleinhuepf:
Awesome, thanks for spotting this Elissavet @elsandal ! And welcome on the list of contributors :-)

PR #162, opened 2022-01-29
Fix Otsu thresholding bin centers

cavenel:
Add minimum_intensity to bin centers' range to get correct values when the minimum intensity of the input image is not zero.


haesleinhuepf:
Awesome, thanks Christophe @cavenel and welcome on the list of contributors! 😃

cavenel:
You're welcome, and thanks for fixing the tests!

PR #165, opened 2022-02-03
Deskew in X

pr4deepr:
Hi  Robert @haesleinhuepf 
I thought I'll try deskew_x and see if it works on some of our data before tackling https://github.com/clEsperanto/pyclesperanto_prototype/pull/164

I think there is some weird scaling going on as the data doesn't appear correct. 
Could you check the code?

Cheers
Pradeep


haesleinhuepf:
Hi @pr4deepr ,

I'll take a look. Furthermore, who's the author of [45_CamA_ch0_stack0006_488nm_0016799msec_0013438399msecAbs-1.tif](https://github.com/clEsperanto/pyclesperanto_prototype/pull/165/files#diff-0e6632f717091819e6ba5ca9416b8c9e422376a7544d61f7b661a2c1c8b27a7f)? Shall we mention them in a small text file next to the dataset?

pr4deepr:
I mentioned it in the notebook.. will add details in a text file too..

haesleinhuepf:
> I think there is some weird scaling going on as the data doesn't appear correct.
Could you check the code?

I've just checked with the other dataset. If I understand the imaging right, we can take the example data from the `deskew_y`-notebook and call `transpose_xy` on it. The resulting dataset is a perfect dataset for testing `deskew_x`. The results in the notebook look ok:

https://github.com/clEsperanto/pyclesperanto_prototype/blob/9764cb3165538b530f701337419f4941190674f6/demo/transforms/deskew_x_with_other_data.ipynb

Hence, I would carefully conclude that there is something wrong with the voxel size of the new test dataset...

What do you think?

PR #168, opened 2022-02-10
Add object-splitting-otsu-labeling

haesleinhuepf:
Hi Jan @biovoxxel ,

I just added the labeling algorithm we were talking about recently, I think it's not exactly similar to your implementation, but close. It aims to be more similar to Voronoi-Otsu-Labeling while still having your idea for the iterative erosion inside. 

You can see how it roughly performs in [this notebook](https://github.com/clEsperanto/pyclesperanto_prototype/blob/d5ab2632fccfb981044d9daa27dcab032a067cd4/demo/segmentation/object_splitting_otsu_labeling.ipynb) and the actual implementation in python lives [here](https://github.com/clEsperanto/pyclesperanto_prototype/blob/d5ab2632fccfb981044d9daa27dcab032a067cd4/pyclesperanto_prototype/_tier9/_object_splitting_otsu_labeling.py#L46-L58).

May I ask you for two things?
* Could you please have a look and let me know if I didn't do something terribly wrong? ;-)
* Let me know what you think about the name of the function! I'm happy to give it a more "griffig" name, if you have one in mind.

Big thanks for the inspiration! 🙌 

Best,
Robert


biovoxxel:
Code looks good as far as I can follow it. Will still need to look at the notebook.
I will also think about the title.

- eroded_otsu_label_splitting

Could at least be a descriptive naming

haesleinhuepf:
Ok, I just renamed it to `eroded_otsu_labeling`. It's also shorter and I like the name 🙌

PR #179, opened 2022-04-26
Shear factor new

pr4deepr:
Hi @haesleinhuepf & @DrLachie 

   I've defined a function `shear_angle_to_shear_factor` so that it calculates the shear factor while taking into account the voxel sizes in the lateral and axial direction.  I'm not sure if this is the best way to describe it. 

Also, I've made an attempt at defining the `deskew_x` function as well. 

Let me know what you think.

Cheers
Pradeep


DrLachie:
Thanks Pradeep, I've been meaning to get back to this for a while. 
I'm not sure about changing to "lateral" and "axial" voxels, just because it's simply XY and Z everywhere else isn't it? Although I guess it makes sense to make the function usable for the general case of X or Y deskewing without confusing the directions, or with non-square voxels?

pr4deepr:
Yea, as you suggested, we would like to perform shearing in any plane, xy, yz, xz, yx etc, so we would like to generalise the terms used. For example, 
for plane yz, y would be the lateral and z the axial; 
for xy plane,  x is lateral and y is axial. 

What about `horizontal axis` and `vertical axis` for each plane?

haesleinhuepf:
Hi @pr4deepr , hi @DrLachie ,

> Yea, as you suggested, we would like to perform shearing in any plane, xy, yz, xz, yx etc, so we would like to generalise the terms used.

Just to be sure I understand correctly: Are there any microscopes that acquire skewed data, where the acquired planes are not stored as Z-planes in the raw data?

pr4deepr:
@haesleinhuepf 

> Just to be sure I understand correctly: Are there any microscopes that acquire skewed data, where the acquired planes are not stored as Z-planes in the raw data?

Not that I know of. They are usually stored as Z planes. I assume you meant how normal microscopy data is stored in XYZ format, right?

I was making a generalised statement for anyone to use this library regardless of data being microscopy or not. 
Does that make sense?

haesleinhuepf:
Ok, cool @pr4deepr! I will happily merge it if the tests pass and the deskweing notebooks work. I just wanted to make sure that you don't spend a lot of effort into something that may never be used ;-) If you think it will be useful, I trust you.

Thanks for your efforts! :sun:

pr4deepr:
I've had to revert the changes I made to shear factor calculation and I've only modified the `deskew_x `function. 
The change in shear factor calculations works for deskewing in y and x, but it was giving incorrect shearing in the ["shearing notebook"](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/demo/transforms/shearing.ipynb). 
As far as I can tell it gives it give accurate results for deskewing in x and y. I've reran the notebooks as well.

haesleinhuepf (approved):
LGTM, big thanks @pr4deepr !

PR #182, opened 2022-06-17
round shape instead before convert to int

pr4deepr:
Hey Robert
      There was a shape mismatch between output from pyclesperanto and Zeiss deskewing. This is because of a bug when we determine the size of the new deskewed image.  We need to round it before converting the shape values to an int. 
Otherwise, if shape was calculated to be 1157.8 in Y dimension, then converting to int will make it 1157 instead of 1158. 
Current fix will use np.around to round the values.

Cheers
Pradeep

### DIFF
--- a/pyclesperanto_prototype/_tier8/_affine_transform.py
+++ b/pyclesperanto_prototype/_tier8/_affine_transform.py
@@ -154,7 +154,7 @@ def _determine_translation_and_bounding_box(source: Image, affine_transformation
     min_coordinate = transformed_bounding_box.min(axis=0)
     max_coordinate = transformed_bounding_box.max(axis=0)
     # determine the size of the transformed bounding box
-    new_shape = (max_coordinate - min_coordinate)[0:3].astype(int).tolist()[::-1]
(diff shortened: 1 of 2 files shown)


haesleinhuepf (approved):
Hey Pradeep @pr4deepr ,

great to hear from you! I hope you are doing well. That proposed change looks great!

Do you see any other issues, or can I just merge this?

Thanks!
Robert

pr4deepr:
Thanks @haesleinhuepf ! Getting back into things and learning to better manage my time.. 😉 
I think this change can be merged. I'll keep monitoring this to make sure this implementation matches with other software. 
Hope you are well!

Cheers
Pradeep

pr4deepr:
I can see that you've approved the changes..
But, just checking, can I merge these changes?

haesleinhuepf:
Hey Pradeep @pr4deepr ,

you're welcome! I'll cut a new release in a minute. 

Greetings from San Francisco!

pr4deepr:
Hope you're having a good time!

PR #204, opened 2022-09-04
Dask arr support and linear interpolation

pr4deepr:
I've added support for resource-backed-dask-array

- [ ] Successfully push a resource-backed-dask-array filetype to the GPU successfully. 
- [ ] Succesfully run a cle operation, such as deskewing on resource-backed-dask-array filetype

TEST CODE

```
# This throws an error in current version of pyclesperanto

from aicsimageio.readers import bioformats_reader

tiff_file = bioformats_reader.BioformatsReader("E:\\RBC_lattice.tif")
print(type(tiff_file.dask_data))



import pyclesperanto_prototype as cle 
import dask.array as da 

tiff_file_3d = da.squeeze(tiff_file.dask_data)
#cle deskewing operation
deskewed = cle.deskew_y(tiff_file_3d,angle_in_degrees=30)
deskewed

```
**************
I've taken this opportunity to also make linear interpolation True by default for deskewing

### DIFF
--- a/pyclesperanto_prototype/_tier0/_types.py
+++ b/pyclesperanto_prototype/_tier0/_types.py
@@ -6,4 +6,4 @@ from typing import Union
 Image = Union[np.ndarray, OCLArray, cl.Image, _OCLImage]
 
 def is_image(object):
-    return isinstance(object, np.ndarray) or isinstance(object, tuple)  or isinstance(object, list) or isinstance(object, OCLArray) or str(type(object)) in ["<class 'cupy._core.core.ndarray'>", "<class 'dask.array.core.Array'>", "<class 'xarray.core.dataarray.DataArray'>"]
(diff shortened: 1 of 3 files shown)


haesleinhuepf (approved):
Hi Pradeep @pr4deepr ,

awesome, thanks!

I just changed the linear-interpolation on your branch so that it's configurable and `False` per default. That goes in line with the other transforms: rotate/scale/... have the same `linear_interpolation:bool = False` parameter ([e.g. here](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/pyclesperanto_prototype/_tier8/_rotate.py#L35)). Through this change, also the tests pass again.

If you see no issues, I will merge it. Let me know what you think. 

Best,
Robert

pr4deepr:
Yup, that makes sense to make it configurable..
Didn't realize it was the case for other tranforms.. Good to know..

haesleinhuepf:
Awesome. Thanks again Pradeep!

PR #224, opened 2022-11-17
Interpolation for deskewing

### LINKED ISSUE #196: Artifact when deskewing
Hi @pr4deepr ,

I just got a screenshot (X-Z view and X-Y view) from a collaborator who sees artifacts when deskewing their data:

image

They saw this artifact in case they had a long Z-slice distance while acquisition. This should be reproducible with our example data as well if we sub-sample in acquisition-Z. 

The issue is that we need to rethink how the interpolation in Z is done. Example:

* This is a raw image on hard-drive:
```
0 0 0 0 1 0 0 0 0 
0 0 0 0 1 2 1 0 0
0 0 0 0 0 0 1 0 0 
```

* But in _reality_, the pixels were acquired like this:
```
    0 0 0 0 1 0 0 0 0 
  0 0 0 0 1 2 1 0 0
0 0 0 0 0 0 1 0 0 
```

Thus, the neighbors of pixel 2 are all supposed to be 1, and not 0. We are interpreting the data wrongly when interpolating.

pr4deepr:
Hi @haesleinhuepf 

  Here comes the pull request for fixing interpolation while deskewing. 
There are two opencl files, 
* [Orthogonal interpolation based on the thesis and on OPM library](https://github.com/clEsperanto/pyclesperanto_prototype/blob/orthogonal_interpolation/pyclesperanto_prototype/_tier8/affine_transform_3d_orth_interp_deskew_y_x.cl)
* two other kernels, [bilinear ](https://github.com/clEsperanto/pyclesperanto_prototype/blob/orthogonal_interpolation/pyclesperanto_prototype/_tier8/affine_transform_3d_interpolate_linear_x.cl)and [trilinear](https://github.com/clEsperanto/pyclesperanto_prototype/blob/orthogonal_interpolation/pyclesperanto_prototype/_tier8/affine_transform_3d_interpolate_trilinear_x.cl) interpolation by finding the inverse transform of the neighbours in the deskewed image. 

Cheers
Pradeep

closes #196


haesleinhuepf (changes requested):
Hi @pr4deepr ,

big thanks again for this PR. Before merging, I would like to clean up things a bit. See my comments below. If you're busy, just let me know what you would do and I do it.

Thanks again!

Best,
Robert

pr4deepr:
Hi Robert
  I'll remove the redundant functions and keep the orthogonal interpolation one. Will address your comments in there as well.  I can work on this, just give me a couple of days. However, if I'm struggling to find time, I'll let you know. 

We will also need an OpenCL kernel for deskew_x with orthogonal interpolation. Will add that option too. 

Apart from the name of the kernel, does `affine_transform_3d_orth_interp_deskew_y` kernel look good? i.e., the input parameters. I'll reference the paper Doug mentioned in the kernel file too.


Cheers
Pradeep

haesleinhuepf:
That sounds awesome @pr4deepr !

> `affine_transform_3d_orth_interp_deskew_y`

Consider calling it `affine_transform_deskew_3d` to keep it simple and readable.

Big thanks again!

pr4deepr:
I've made changes as requested and only kept the kernels for orthogonal interpolation. 
**There are two kernels. One for deskew_x and one for deskew_y.**.
 They could be merged, but that may make the opencl kernel too long and have too many if statements. I think this is cleaner,especially for finding any bugs. 

- I've added an option for[ `flip_z` ](https://github.com/clEsperanto/pyclesperanto_prototype/blob/67658dcd875d9219857cda431a9ca73313c28bb8/pyclesperanto_prototype/_tier8/_deskew_y.py#L14)for coverslip rotation in deskew function and in the kernel [`rotate_bool`](https://github.com/clEsperanto/pyclesperanto_prototype/blob/67658dcd875d9219857cda431a9ca73313c28bb8/pyclesperanto_prototype/_tier8/affine_transform_deskew_y_3d_x.cl#L44). Maybe rotate_bool could be changed to a better name?
- I've fixed a bug in the kernel and now it takes three new parameters as input, i.e.,  [the shape of the deskewed image](https://github.com/clEsperanto/pyclesperanto_prototype/blob/67658dcd875d9219857cda431a9ca73313c28bb8/pyclesperanto_prototype/_tier8/affine_transform_deskew_y_3d_x.cl#L45). 
- I've reran the notebooks and it seems to be working great. At the end of the [deskew notebook](https://github.com/clEsperanto/pyclesperanto_prototype/blob/79e2e8c7057ede5abcb028ea38004a17796265be/demo/transforms/deskew.ipynb), I've added an example to check the interpolation (XZ view). 
- For the interpolation, I've added references to Doug's OPM library, Vincents' thesis and the elife paper in the OpenCL kernel and the deskew notebook. I think we should also have it in the release notes.

Cheers
Pradeep

pr4deepr:
So, I'm trying to write tests for deskew. 
I was comparing OPM and pyclesperanto and it is giving different results. Not sure if this is a GPU vs CPU implementation detail or difference in algorithm? As far as I can tell, the algorithm is similar. Only thing I can think of is whether the types of the variables I've initialized in the OpenCL kernel are similar to the OPM one?

I've added a notebook comparing the differences using a simple 10 x 10 x 10 array and using test data.

pr4deepr:
I think I've fixed it. 
Within the OpenCL kernel for deskew_y and deskew_x, I was using float type for coordinate, instead of int. As we are not using normalized coordinates, I changed it to uint type and that seems to work. 
_You can view the important changes in this [**commit**]_(https://github.com/clEsperanto/pyclesperanto_prototype/pull/224/commits/7c4b5005676b5b42256fee0b203e91fd12521edd). 
I've updated the tests to reflect this.

If the tests are successful, I think it should be good to go. Let me know what you think..

haesleinhuepf (approved):
LGTM

Big thanks again @pr4deepr !

PR #230, opened 2022-11-25
Added maximum position function

### LINKED ISSUE #188: cle.maximum_position
As discussed in #185 , it would be cool to have a function `maximum_position` similar to the original in scipy:
* https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.maximum_position.html

I _think_ we could potentially reuse code in [ZPositionOfMaximumZProjection](https://github.com/clij/clij2/blob/master/src/main/java/net/haesleinhuepf/clij2/plugins/ZPositionOfMaximumZProjection.java) and potentially make similar functions for X and Y. 

Before starting this, it might be a good idea to clearly define the behaviour of this function, e.g. in a 3D image that has multiple maxima.

grahamross123:
Here's my implementation of a maximum position function (similar to [ndi.maximum_position](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.maximum_position.html)) that returns the coordinates of the maximum value in an array.

It works similar to `maximum_of_all_pixels` but it also tracks the coordinates with `x_position_of_maximum_x_projection` and `y_position_of_maximum_y_projection`.

To have the same function as `ndi.maximum_position`, projections are found in the order X -> Y -> Z (instead of Z -> Y -> X as with maximum_of_all_pixels)

This probably needs a bit more optimisation and some tests still have to be written.

closes #188


haesleinhuepf (approved):
LGTM. Thanks @grahamross123 !

PR #232, opened 2022-12-01
Pytorch interoperability

haesleinhuepf:
Hi @alisterburt,

I've just added interoperability with torch tensors. You can see how it works in [this notebook](https://github.com/clEsperanto/pyclesperanto_prototype/blob/a45949e7bbacab4d9b8fd9f2e9f19715442877f5/demo/interoperability/torch.ipynb). Does that fit to your use-case mentioned in the [image.sc forum](https://forum.image.sc/t/connected-component-transform-in-python/74563/6?u=haesleinhuepf). Let me know if there is anything missing!

Thanks!

Best,
Robert


alisterburt:
Beautiful! fantastic stuff, and so fast!! Thanks :-)

PR #233, opened 2022-12-01
Deskew version 2

haesleinhuepf:
Hi @pr4deepr ,

I'm proposing here some changes to your updated deskewing code:
* I'm removing `affine_transform_deskew_3d` from the public API as I don't think this is a function ready for end-users. If there is a need for having it publicly accessible, let's discuss.
* I'm removing the 'flip_z' / 'rotate' parameter for two reasons: 
  * A parameter that flips the image should not be called "rotate". 
  * If users want to flip the image, they can use the `cle.flip` function. We should not have _functional_ parameters in functions. The end-point of this strategy would be that every function has a `flip_z` parameter in case the user wants to flip the image. There would also be parameters `flip_x` and `flip_y`, and `transpose_xy` and ... in aaall functions ;-) 
* I renamed the parameter 'skew_direction' to 'deskew_direction'. Otherwise it would imply there is a difference between the skew-direction and the deskew-direction.
* For the `deskew_direction` I'm now using an `Enum` and not a `str`. Using strings for this is python2 style. Enums are the right thing to use here. When using Enums, the python interpreter can point you at typos. When you use strings and have a typo in there, issues are hard  to find.
* I added back the `linear_interpolation` parameter to the `deskew_x`/`y` function and added a warning that this parameter is obsolete. If we removed this parameter without further notice, users of the parameter would experience a crash after updating pyclesperanto, because we broke backwards-compatibility. A warning is more gentle. Btw. if possible, never break backwards compatibility.

If you're ok with these changes, I would merge your new code and my modifications together into master and release a new version.

Big thanks again! That's an affine masterpiece!

Best,
Robert


pr4deepr:
Thanks for the explanation @haesleinhuepf 
Learned quite a few things, especially ensuring backwards compatibility..

Yes, I'm happy with the changes and agree with your points. 

Please go ahead and merge all of this!

Thanks for your help with this.

Cheers
Pradeep

pr4deepr (approved):
Thanks. 
This looks much better and I agree with not having functional parameters in functions.. 
Although, I did really like how I added flip within the OpenCL kernel.. 😆 
Use of Enums is a much better idea too. I'll be using these ideas in other repos too!

haesleinhuepf:
> Although, I did really like how I added flip within the OpenCL kernel.. 😆

It was indeed very efficiently integrated. Just be aware that the OpenCL-compiler typically does these things better than we do. If you build a pipeline where deskewing and flipping are subsequent operations, it will generate compiled  GPU-code (assembly) that does pretty much what you did, potentially more efficiently, depending on compiler and hardware. 

Big thanks again!

PR #244, opened 2023-01-12
bugfix for linux

ENicolay:
as just discussed


haesleinhuepf:
Awesome, thanks Elena @ENicolay, it was fun debugging this!

PR #249, opened 2023-01-25
Updated Affine interpolation

iionichi:
I have updated the affine transform kernel affine_transform_2d_x.cl and added cle_test.ipynb with the test we did yesterday (Tuesday). Please let me know how well it works. The mean square error (for rotation) is quite small so I am assuming it works fine.


haesleinhuepf:
Awesome, thanks @iionichi !

PR #267, opened 2023-02-27
added standard error of the mean to statistics_of_labelled_pixels

thawn:
I think this should calculate the SEM, but please double check. If I read the code correctly, this automagically adds SEM also to `statistics_of_background_and_labelled_pixels`

### DIFF
--- a/pyclesperanto_prototype/_tier9/_statistics_of_labelled_pixels.py
+++ b/pyclesperanto_prototype/_tier9/_statistics_of_labelled_pixels.py
@@ -276,6 +276,11 @@ def statistics_of_labelled_pixels(intensity_image : Image = None, label_image : 
     power(sum_dim, result_vector, 0.5)
     region_props['standard_deviation_intensity'] = pull(result_vector)[0]
 
+    # standard error intensity
+    crop(sum_per_label, sum_dim, measurements_start_x, 3, 0) # number of pixels/voxels = area
+    power(sum_dim, result_vector, 0.5)
+    region_props['standard_error_intensity'] = region_props['standard_deviation_intensity'] / pull(result_vector)[0]
+
     crop(max_statistics, result_vector, measurements_start_x, 4, 0)
     region_props['max_distance_to_centroid'] = pull(result_vector)[0]
     crop(max_statistics, result_vector, measurements_start_x, 5, 0)


pyclesperanto_prototype/_tier9/_statistics_of_labelled_pixels.py
```diff
@@ -276,6 +276,11 @@ def statistics_of_labelled_pixels(intensity_image : Image = None, label_image :
     power(sum_dim, result_vector, 0.5)
     region_props['standard_deviation_intensity'] = pull(result_vector)[0]
 
+    # standard error intensity
+    crop(sum_per_label, sum_dim, measurements_start_x, 3, 0) # number of pixels/voxels = area
```
haesleinhuepf, line 280: Ok, wow. I must admit, I don't get this. How about using `region_props['area']` ?

thawn: This is area calculation copied and pasted from line 199 (the same also appears on line 260) above.
I am assuming this leaves the area in the GPU memory, so that I can then use `power()` to calculate the square root, rather than importing np.sqrt or math.sqrt just for this.

thawn: How would you calculate the square root of `region_props['area']`?

haesleinhuepf: `np.sqrt(region_props['area'])`? (I'm speculating)

thawn: is this better :-)

actually that was my initial approach, but then I saw that you used power() everywhere else, so I tried to follow your coding style... ;-)

thawn:
I am just working on the tests and a question came up:
should the statistics also appear in the output of push_regionprops()?
if yes, should they be inserted where they belong semantically in pos 15 below props['standard_deviation_intensity'], or where they are backwards compatible at the end (pos 36)?

furthermore, I feel confident, that I can add the code in case the input is a dict. in the case that the input is a RegionProperties object, I am less confident... (and also don't know if the standard error would appear in that case)

thawn:
@haesleinhuepf I now added a simple test for the standard error. Would you like to also add the standard error to push_regionprops (see my question above?)

haesleinhuepf:
> Would you like to also add the standard error to push_regionprops (see my question above?)

Yes, it seems this makes sense! Thank you!

thawn:
@haesleinhuepf I just added the SEM to `push_regionprops`. Also adapted the tests accordingly.

PR #296, opened 2023-04-03
Morph snake smoothing

grahamross123:
Hi, I've added `superior_inferior` and `inferior_superior` functions as used in the smoothing process in scikit-image [morphsnakes](https://github.com/scikit-image/scikit-image/blob/00177e14097237ef20ed3141ed454bc81b308f82/skimage/segmentation/morphsnakes.py).

There's one issue with the tests for both 2D and 3D:

If there are pixels adjacent to the image boundaries, the scikit-image implementation and my implementation differ. I think this is due to the opencl kernel taking pixels outside the image boundary as 1, whereas it should be 0.

You can see this in the final example in the [notebook](https://github.com/clEsperanto/pyclesperanto_prototype/blob/morph_snake_smoothing/demo/smoothing.ipynb).

Is there an easy way to fix this problem in the opencl kernel?

Thanks,
Graham


grahamross123:
@haesleinhuepf Here's the pull request I mentioned!

PR #297, opened 2023-04-06
Dask demo with GPU cluster

jackyko1991:
This demo is for the illustration of pyclespranto integration with Dask.

Update the original dask gaussian blur demo with proper documentation and big data handling.

Unlike Dask LocalCUDACluster, pyclespranto GPU memory is not directly handled by Dask. The trick to work around for a heterogeneous dask local cluster. 

GPU memory cycle when using Dask is also included to reduce memory overflow on parallel processes.


haesleinhuepf:
Awesome, big thanks @jackyko1991 !

PR #304, opened 2023-04-30
Morphological snakes2

haesleinhuepf:
closes #301 
closes #229 
closes #296


pyclesperanto_prototype/_tier3/_morphological_chan_vese.py
```diff
@@ -0,0 +1,170 @@
...
+    greater_constant(contour_image, destination=output_image, constant=0)
+
+    temp_1 = create_like(output_image)
+    temp_2 = create_like(output_image)
+    temp_3 = create_like(output_image)
+    temp_4 = create_like(output_image)
+    temp_5 = create_like(output_image)
+
+    for _ in range(num_iter):
+        set(temp_3, 0)
```
haesleinhuepf, line 84: Hint: I had to initialize this image here with 0 because the `create*` functions only allocate memory. Further down, we compute `temp_3 += ....` which might be dangerous with a non-initialized image.

grahamross123:
Thanks for finishing this @haesleinhuepf! It works as expected for me.

